PR #1, opened 2022-05-18
Feature/rename modules add action

ajfloeder:


Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM golang:1.17 AS builder
 
 # Set Go environment
-ENV GOPRIVATE="github.hpe.com"
+ENV GOPRIVATE="github.com"
```
roehrich-hpe, line 4: github.com is not private.  So we can 
 delete this line  now.

roehrich-hpe (approved):

PR #3, opened 2022-05-25
RABSW-883: User owner labels for child deletion

matthew-richerson:
This commit uses the new owner label code from DWS to apply owner labels
to child resources and use them for deletion.

The changes included in this commit are:
 - Add owner labels to all child resources
 - Add Workflow or PersistentStorageInstance labels to all children
 - Inherit parent labels where appropriate to pass the Workflow and
   PersistentStorageInstance labels through the resource chain
 - Remove ObjectReferences in NnfStorage. This required changes in the
   nnfAccess controller as well. NnfAccess uses the NnfNodeStorages to
   determine client-storage mappings. This is done using the new owner
   labels now as well.
 - Create a PersistentStorageInstance controller that creates/deletes
   the Servers resource, and deletes the NnfStorage resource.
 - All resources are deleted in the correct order when the workflow progresses
   directly to "teardown"
 - All resources are cleaned up properly if the workflow is deleted
   before it reaches "teardown"
 - The servers resource was changed to use a map in the status section for
   tracking existing allocations. This allows better handling of when
   the same Rabbit is listed multiple time in the spec section.
 - The NnfAccess adds a finalizer to the NnfStorage it is using. This is
   a temporary solution for PersistentStorageInstances that can be
   deleted before all workflows that are using them have finished.


controllers/nnf_workflow_controller.go
```diff
@@ -120,22 +122,49 @@ func (r *NnfWorkflowReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			return ctrl.Result{}, nil
 		}
 
-		// TODO: Teardown workflow
-		/*
-			if err := r.teardownWorkflow(ctx, workflow); err != nil {
-				return ctrl.Result{}, err
-			}
+		// Run the teardown state logic in case the workflow was deleted without
```
roehrich-hpe, line 125: We can use the validation webhook in dws-operator to prevent a delete-without-teardown.  Add it to the ValidateDelete() method there.

controllers/nnf_workflow_controller.go
```diff
@@ -1484,155 +1466,69 @@ func (r *NnfWorkflowReconciler) handlePostRunState(ctx context.Context, workflow
...
-		return ctrl.Result{}, err
+		return true, err
 	}
 
-	if exists == resourceExists {
-		return ctrl.Result{}, nil
+	childObjects := []dwsv1alpha1.ObjectList{
+		&nnfv1alpha1.NnfDataMovementList{},
+		&nnfv1alpha1.NnfAccessList{},
+		&nnfv1alpha1.NnfStorageList{},
```
roehrich-hpe, line 1480: The NnfStorageProfile goes here as well.  --Actually, the nnfstorage owns the nnfstorageprofile.

controllers/nnf_storage_controller.go
```diff
@@ -349,67 +328,29 @@ func (r *NnfStorageReconciler) aggregateNodeStorageStatus(ctx context.Context, s
...
-					log.Info("Unable to get NnfNodeStorage", "Error", err, "NnfNodeStorage", namespacedName)
-				}
-			} else {
-				state = nodeStoragesExist
-				if !nnfNodeStorage.GetDeletionTimestamp().IsZero() {
-					continue
-				}
-			}
+	childObjects := []dwsv1alpha1.ObjectList{
+		&nnfv1alpha1.NnfNodeStorageList{},
```
roehrich-hpe, line 341: Add NnfStorageProfile here.  Add your labels to it as well, because it's owned by the NnfStorage.
Look for setPinnedProfileOwnership().

Or are there certain kinds of resources that don't require your new managed deletion?

controllers/nnf_workflow_controller.go
```diff
@@ -1641,34 +1537,15 @@ func (r *NnfWorkflowReconciler) deletePersistentNnfStorages(ctx context.Context,
...
-				storageStatus, firstErr = captureErrorAndStatus(firstErr, err, storageStatus, psiStatus)
-			}
 		}
 	}
 
-	return storageStatus, firstErr
+	return firstErr
 }
 
 func getPersistentName(deleteDirective string) (string, error) {
```
roehrich-hpe, line 1551: This  function can be  deleted  now.

controllers/test
```diff
@@ -0,0 +1 @@
+make: Nothing to be done for `test'.
```
roehrich-hpe, line 1: Why is this file here?

controllers/nnf_workflow_controller.go
matthew-richerson: The workflow will be properly cleaned up even if it isn't in the teardown state, so there's no need to restrict delete to only when we're in teardown.

controllers/nnf_storage_controller.go
matthew-richerson: Nope it should have it too. I just missed it.

controllers/test
matthew-richerson: I'm not sure how that happened. I'll delete it

controllers/nnf_storage_controller.go
roehrich-hpe: Can we pull these lists out and put them, empty, in `api/*/foo_types.go`?  Then populate the lists in each  SetupWithManager().  I'd like to get the lists to a place where they're easy to find and where future devs are more likely to maintain them as new CRDs are created.  Building them in SetupWithManager() puts them in a noticeable place as people are updating their Owns() and Watches().

ajfloeder (approved):

controllers/nnf_storage_controller.go
```diff
@@ -452,33 +391,17 @@ func (s *storageStatusUpdater) close(ctx context.Context, r *NnfStorageReconcile
...
-			Namespace: components[1],
-		}},
-	}
-}
-
 // SetupWithManager sets up the controller with the Manager.
 func (r *NnfStorageReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	r.ChildObjects = []dwsv1alpha1.ObjectList{
+		&nnfv1alpha1.NnfNodeStorageList{},
+		&nnfv1alpha1.NnfStorageProfileList{},
```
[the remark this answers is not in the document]
matthew-richerson: It's in nnf_workflow_controller and nnf_storage_controller since the parent switches from the workflow to the nnfstorage during the setup phase.

PR #12, opened 2022-06-09
RABSW-927: Include compute breakdown for persistentdw

matthew-richerson:
This commit updates sos to use the new layout for the directiveBreakdown that
has a section for storage allocations and a section for compute allocations.
The compute information is filled in with a location constraint for specifying
how to choose compute nodes based on the storage they need access to. The
location constraint provides an ObjectReference to the Servers resource where
the storage is defined, and information on how that storage can be accessed
(physical vs. network).

I also changed the owner reference on the PersistentStorageInstance created by
the directiveBreakdown controller so that it's owned by the DirectiveBreakdown
through the Setup phase. This ensures that it's cleaned up if anything fails
while it's creating.

There were lots of code changes that were a result of the changed DirectiveBreakdown
fields and having a DirectiveBreakdown for persistentdw now.


controllers/directivebreakdown_controller.go
```diff
@@ -122,30 +129,74 @@ func (r *DirectiveBreakdownReconciler) Reconcile(ctx context.Context, req ctrl.R
...
 		// If we failed to create the persistent instance or we created it this pass, requeue.
-		if persistentInstance == nil || result == controllerutil.OperationResultCreated {
-			return ctrl.Result{Requeue: true}, nil
+		if persistentStorage == nil || result == controllerutil.OperationResultCreated {
+			return ctrl.Result{}, nil
 		}
 
-		if persistentInstance.Status.Servers == (v1.ObjectReference{}) {
-			return ctrl.Result{Requeue: true}, nil
+		if persistentStorage.Status.Servers == (v1.ObjectReference{}) {
```
ajfloeder, line 152: Would you please add a comment to ease the burden when I read this again that this is checking whether the Servers is created? 
Also, will this controller wake up when the persistent instance sets Status.Servers because it owns it?

controllers/integration_test.go
```diff
@@ -429,9 +429,9 @@ var _ = Describe("Integration Test", func() {
 			{"#DW create_persistent name=createpersistent-gfs2   type=gfs2   capacity=1GiB", 1, 1},
```
ajfloeder, line 429: Alignment seems off here.

controllers/integration_test.go
```diff
@@ -522,9 +522,7 @@ var _ = Describe("Integration Test", func() {
 				By("Checking DW Directive Breakdown")
```
ajfloeder, line 522: Alignment seems off here too.

ajfloeder, line 522: Please add checking for the dbd.Status.Computes values that you expect here.

controllers/nnf_workflow_controller.go
```diff
@@ -257,8 +257,7 @@ func (r *NnfWorkflowReconciler) handleUnsupportedState(ctx context.Context, work
 
 	log.Info("Unsupported Status.State", "state", workflow.Status.State)
 
-	// TODO: This should return an error, but for now unhandled states simply succeed to allow Flux to exercise workflow...
-	// Complete state in the drivers
+	// TODO: This should return an error, but for now unhandled states simply succeed to allow Flux to exercise worworkflow.Spec.DWDirectives[dwIndex]	// Complete state in the drivers
```
ajfloeder, line 260: Does this build? Where is worworkflow defined?
I think we handle all states now, so the comment is defunct.

controllers/nnf_workflow_controller.go
```diff
@@ -489,7 +498,7 @@ func (r *NnfWorkflowReconciler) getPersistentStorageInstance(ctx context.Context
 }
 
 // generateDirectiveBreakdown creates a DirectiveBreakdown for any #DW directive that causes storage to be allocated.
```
ajfloeder, line 500: Persistentdw also causes a dbd, need to update this comment.

controllers/nnf_workflow_controller.go
```diff
@@ -499,30 +508,27 @@ func (r *NnfWorkflowReconciler) generateDirectiveBreakdown(ctx context.Context,
 	// "#DW jobdw              type=gfs2   capacity=9GB name=thisIsGfs2"
 	// "#DW jobdw              type=lustre capacity=9TB name=thisIsLustre"
 	// "#DW create_persistent  type=lustre capacity=9TB name=thisIsPersistent"
+	// "#DW persistentdw       name=thisIsPersistent"
 
 	// #DW commands that require a dwDirectiveBreakdown because they create storage
```
ajfloeder, line 513: Comment is stale. persistentdw needs to specify which computes can be chosen for the WLM.

ajfloeder (commented):
A couple of the files have odd indentation in this tool. Have you run go fmt?

controllers/integration_test.go
matthew-richerson: I'm not seeing the alignment issue. You're talking about like 429?

controllers/integration_test.go
matthew-richerson: I'll handle this in a separate PR.

matthew-richerson: I'm not seeing this alignment issue either.

controllers/nnf_workflow_controller.go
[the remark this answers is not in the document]
matthew-richerson: I must have. This wasn't intentional. I'll fix it.

matthew-richerson: I'll leave the comment for now. We need to remove the code that was treating unhandled states as success. I'll save that for a different PR.

ajfloeder (approved):

roehrich-hpe (approved):

PR #19, opened 2022-06-24
RABSW-956: Change pinned profile ownership

matthew-richerson:
This commit changes the ownership of the pinned profiles. Pinned profiles
are owned by either:
 - The DirectiveBreakdown for jobdw
 - The PersistentStorageInstance for create_persistent

The pinned profiles do not change ownership after they've been created.

I had to disable a few of the storage profile tests because we don't have the path
to bring errors from the DirectiveBreakdown into the Workflow.

### DIFF
--- a/controllers/directivebreakdown_controller.go
+++ b/controllers/directivebreakdown_controller.go
@@ -41,6 +41,7 @@ import (
 
 	dwsv1alpha1 "github.com/HewlettPackard/dws/api/v1alpha1"
 	"github.com/HewlettPackard/dws/utils/dwdparse"
+	nnfv1alpha1 "github.com/NearNodeFlash/nnf-sos/api/v1alpha1"
 )
 
 // Define condition values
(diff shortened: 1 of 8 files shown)


NateRoiger:
Is the failure we see https://github.com/NearNodeFlash/nnf-sos/runs/7042378727?check_suite_focus=true#step:8:2887 something that occurs naturally? Or did this expose/introduce a bug?

matthew-richerson:
> Is the failure we see https://github.com/NearNodeFlash/nnf-sos/runs/7042378727?check_suite_focus=true#step:8:2887 something that occurs naturally? Or did this expose/introduce a bug?

That's the test that fails occasionally. I'll kick it off again and see if it's successful.

roehrich-hpe (commented):
In nnf_workflow_controller.go, do we still need the Owns(NnfStorageProfile)?

matthew-richerson:
> In nnf_workflow_controller.go, do we still need the Owns(NnfStorageProfile)?

nope I'll remove it

PR #25, opened 2022-07-08
RABSW-970: Process directives individually

matthew-richerson:
This commit changes the layout of the workflow controller to allow finer grained control of
which directives are processed, and how the success and failure of that directive is reported
in the workflow. It changes the "handler" functions to process a single directive specified
by its index in the workflow. This is different from before since the "handler" functions would
process all the directives in the workflow before marking them as successful.

The handler functions were changed to have a separate "start" and "finish" handler for each
workflow state (e.g., startProposalState and finishProposalState). This makes it easy to code
for adding parallelism when there are multiple directives. The "start" function creates all the
child objects, and the "finish" function waits for the child resource to become ready.

Which directives to process was standardized. Before we were doing a mixture of parsing the directive
list, walking the directive breakdowns, and looking at the driver status array in the workflow.
Everything is now based on the information in the driver status array of the workflow. This has
all the information of which directives are interested in which workflow states. It also allows us
to skip over directives that have already completed the current state.


controllers/nnf_workflow_controller_test.go
```diff
@@ -63,7 +63,10 @@ var _ = Describe("NNF Workflow Unit Tests", func() {
 
 	AfterEach(func() {
 		Eventually(func() error {
-			Expect(k8sClient.Get(context.TODO(), key, workflow)).To(Succeed())
```
NateRoiger, line 66: I think the correct way to do this is to add an input parameter `g Ginkgo` and change the Expect to `g.Expect`

Sample from nnf-dm
```go
Eventually(func(g Gomega) dmv1alpha1.RsyncNodeDataMovementStatus {
	g.Expect(k8sClient.Get(context.TODO(), client.ObjectKeyFromObject(rsync), rsync)).To(Succeed())
	return rsync.Status
})
```

matthew-richerson: Right. I'll change it

roehrich-hpe: Yes, please.
It's described in this section of the Gomega  doc: https://onsi.github.io/gomega/#eventually

controllers/nnf_workflow_controller.go
```diff
@@ -1098,220 +1011,269 @@ func (r *NnfWorkflowReconciler) setupNnfAccessForServers(ctx context.Context, st
...
-					return ctrl.Result{}, err
-				} else if result == controllerutil.OperationResultCreated {
-					log.Info("Created NnfAccess", "name", access.Name)
-				} else if result == controllerutil.OperationResultUpdated {
-					log.Info("Updated NnfAccess", "name", access.Name)
-				}
+	for _, suffix := range accessSuffixes {
+		access := &nnfv1alpha1.NnfAccess{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      fmt.Sprintf("%s-%d-%s", workflow.Name, index, suffix),
```
NateRoiger, line 1240: `indexedResourceName(workflow, index) + suffix` ?

controllers/nnf_workflow_controller.go
```diff
@@ -1076,6 +985,10 @@ func (r *NnfWorkflowReconciler) setupNnfAccessForServers(ctx context.Context, st
 		func() error {
 			dwsv1alpha1.AddWorkflowLabels(access, workflow)
 			dwsv1alpha1.AddOwnerLabels(access, workflow)
+			labels := access.GetLabels()
+			labels[teardownStateLabel] = teardownState
+			labels[directiveIndexLabel] = strconv.Itoa(directiveIdx)
+			access.SetLabels(labels)
 
 			access.Spec = nnfv1alpha1.NnfAccessSpec{
 				DesiredState:    "mounted",
```
NateRoiger, line 994: Can you also fix line 997 to be buildMountPath()

controllers/nnf_workflow_controller.go
```diff
@@ -1547,17 +1560,27 @@ func (r *NnfWorkflowReconciler) findPersistentInstance(ctx context.Context, wf *
...
+		existingStatus: (*w.DeepCopy()).Status,
+	}
+}
 
-	if err := controllerutil.SetControllerReference(wf, psi, r.Scheme); err != nil {
-		log.Error(err, "Unable to assign workflow as owner of persistentInstance")
-		return err
+func (w *workflowStatusUpdater) close(ctx context.Context, r *NnfWorkflowReconciler) error {
+	if !reflect.DeepEqual(w.workflow.Status, w.existingStatus) {
+		err := r.Update(ctx, w.workflow)
```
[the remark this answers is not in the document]
matthew-richerson: The workflow doesn't have status as a sub-resource, so the status has to be updated with r.Update(). From what I remember, that was due to the mutating web hook needing to add the driver array entries to the status section.

NateRoiger:
Excellent refactor - thanks for doing this

controllers/nnf_workflow_controller.go
```diff
@@ -1547,17 +1546,27 @@ func (r *NnfWorkflowReconciler) findPersistentInstance(ctx context.Context, wf *
...
+		if !apierrors.IsConflict(err) {
```
roehrich-hpe, line 1564: If this is a conflict, then you'll return nil...will we get the retry?
Maybe this should return a result that asks for a retry.

NateRoiger: Seems true for all our `statusUpdater`s. Maybe address this in another PR?

(I've been wanting to jump to go 1.18 and make status updater a generic)

roehrich-hpe:
I'm unclear on how destroy_persistent progresses through this.  Do your changes prevent it from completing the proposal state?  Is it supposed to progress through each state?

controllers/nnf_workflow_controller.go
roehrich-hpe: If I'm right, then in this instance it can't wait for another PR.  The first conflict will stop the workflow from progressing, I think.

roehrich-hpe: I won't be finished with this PR before I stop for the day.  I'd like to have more time on Monday to continue reading through it.

matthew-richerson:
> I'm unclear on how destroy_persistent progresses through this. Do your changes prevent it from completing the proposal state? Is it supposed to progress through each state?

I'm not sure I follow your question. destroy_persistent only registers for the proposal state and the teardown state. The proposal state just goes through the validation to make sure that there aren't other #DWs other that destroy_persistent. (I think it would be nice to move all that verification code from the proposal state to a validating web hook). Teardown state is where the real work happens to delete the persistentStorageInstance.

controllers/nnf_workflow_controller.go
matthew-richerson: That's fine with me. I'm wrapping up for the day now.

roehrich-hpe (approved):

controllers/nnf_workflow_controller.go
```diff
@@ -206,132 +193,182 @@ func (r *NnfWorkflowReconciler) updateDriversStatusForStatusState(workflow *dwsv
...
-	// Complete state in the drivers
-	return r.completeDriverState(ctx, workflow, driverID, log)
+	return ctrl.Result{}, nil
 }
 
-func (r *NnfWorkflowReconciler) handleProposalState(ctx context.Context, workflow *dwsv1alpha1.Workflow, driverID string, log logr.Logger) (ctrl.Result, error) {
-	log.Info(workflow.Status.State)
+func (r *NnfWorkflowReconciler) startProposalState(ctx context.Context, workflow *dwsv1alpha1.Workflow, index int) (*ctrl.Result, error) {
+	log := r.Log.WithValues("Workflow", client.ObjectKeyFromObject(workflow), "Index", index)
+	dwArgs, _ := dwdparse.BuildArgsMap(workflow.Spec.DWDirectives[index])
```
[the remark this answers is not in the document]
matthew-richerson: I started doing this and then decided maybe it's not an improvement. Not all of the "start"/"finish" functions need the dwArgs, so by putting it in the caller we're doing unnecessary work in a lot of cases. It's only one line of code, and doing it this way keeps the argument list shorter.

controllers/nnf_workflow_controller.go
```diff
@@ -1098,220 +1003,265 @@ func (r *NnfWorkflowReconciler) setupNnfAccessForServers(ctx context.Context, st
...
 
-		// Parse the #DW line
-		args, err := dwdparse.BuildArgsMap(workflow.Spec.DWDirectives[driverStatus.DWDIndex])
-		if err != nil {
-			return ctrl.Result{}, err
-		}
+	// Create an NnfAccess for the servers resources if necessary. Shared storage like
+	// that of GFS2 provides access to the Rabbit for the lifetime of the user's job.
+	// NnfAccess may already be present if a data_in directive was specified for the
+	// particular $JOB_DW_[name]; in this case we only don't need to recreate the
```
ajfloeder, line 1173: "we only don't need" ??

initiateContainerTest.sh
```diff
@@ -21,6 +21,8 @@ CWD=`pwd`
 echo "Running in $CWD, setting up envtest."
 
 export ENVTEST_ASSETS_DIR=/nnf/testbin
+export GOMEGA_DEFAULT_EVENTUALLY_TIMEOUT=20s
+export GOMEGA_DEFAULT_EVENTUALLY_INTERVAL=100ms
```
roehrich-hpe, line 25: Given the volume of output from this particular test suite, I'd like to see the following added to the "go test" commandline below:  `-args -ginkgo.v -ginkgo.progress -ginkgo.failFast`
We use that in the  makefile.

roehrich-hpe (approved):

initiateContainerTest.sh
roehrich-hpe: Bah, I didn't mean to click approve with that...

controllers/nnf_workflow_controller.go
```diff
@@ -53,6 +55,17 @@ const (
...
 	// reconciler has finished in using the resource.
 	finalizerNnfWorkflow = "nnf.cray.hpe.com/nnf_workflow"
+
+	// directiveIndexLabel is a label applied to child objects of the workflow
+	// to show which directive they were created for. This is useful during deletion
+	// to filter the child objects by the directive index and only delete the
+	// resources for the directive being processed
+	directiveIndexLabel = "nnf.cray.hpe.com/directive_index"
+
+	// teardownStateLabel is a label applied to NnfAccess resources to determine when
```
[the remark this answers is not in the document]
matthew-richerson: I think it's only useful for the NnfAccesses. Depending on the combination of #DWs in a workflow, the NnfAccess resources get deleted in different states. The other child resource types don't have that issue.

controllers/nnf_workflow_controller.go
```diff
@@ -206,132 +193,182 @@ func (r *NnfWorkflowReconciler) updateDriversStatusForStatusState(workflow *dwsv
...
+				} else {
+					driverStatus.Reason = "error"
+				}
+			} else {
+				driverStatus.Reason = "error"
+				driverStatus.Message = "Internal error: " + err.Error()
 			}
 
-			log.Error(err, "Failed to update Workflow state")
+			log.Info("Start error", "State", workflow.Status.State, "index", driverStatus.DWDIndex)
```
roehrich-hpe, line 232: Log the error here.   I'm hitting this line in my test runs, and I'm having trouble figuring out where the error message is in the test output.

roehrich-hpe (approved):

roehrich-hpe:
Did this require the DWS PR https://github.com/HewlettPackard/dws/pull/18 ?

matthew-richerson:
> Did this require the DWS PR [HewlettPackard/dws#18](https://github.com/HewlettPackard/dws/pull/18) ?

Yeah, I forgot to do that one first. I'll update this repo with another PR to vendor DWS from master

PR #37, opened 2022-07-21
NNF Element Controller Recovery

NateRoiger:
Introduces a new CRD, NnfNodeECData (sorry Tony) for holding the critical data for nnf-ec. The corresponding controller implements the StorageProvider interface.

Had to reshuffle where nnf-ec starts to ensure that the StorageProvider (the reconciler) is assigned before nnf-ec.Init(). This caused a couple changes to the NnfNode and NnfNodeStorage so they delay in accessing the NNF Storage Service until it is ready. 

I also retired the service/endpoint from NnfNode code - nobody uses the service name anyways and it just complicated things.

Signed-off-by: Nate Roiger <nate.roiger@hpe.com>


ajfloeder (approved):

ajfloeder (approved):

controllers/nnf_node_ec_data_controller.go
```diff
@@ -0,0 +1,205 @@
...
+			}
+
+			if err := r.Create(ctx, &data); err != nil {
+				return err
+			}
+		}
+	}
+
+	// This resource acts as the storage provider for the NNF Element Controller
+	persistent.StorageProvider = r
```
matthew-richerson, line 82: Do you need the whole reconciler here, or just the client?

controllers/nnf_node_ec_data_controller.go
```diff
@@ -0,0 +1,205 @@
+/*
```
matthew-richerson, line 1: It seems weird to me that this code is in nnf-sos and is part of the controller framework. It's not a reconciler, and it's not part of nnf-sos. Is there something preventing you from making a k8s client down in nnf-ec? Or passing the existing k8s client from nnf-sos to nnf-ec if you wanted to re-use it?

matthew-richerson (approved):

controllers/nnf_node_ec_data_controller.go
```diff
@@ -0,0 +1,205 @@
...
+
+		return err
+	}
+
+	return nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *NnfNodeECDataReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	if err := mgr.Add(r); err != nil {
```
roehrich-hpe, line 198: Add a comment like this one that you have in nnf_node_controller:
```// Causes NnfNodeReconciler.Start() to be called.```
Sometimes these clues help to stitch things together;  in this case it takes a bit to work it out even with the help of VSCode.

controllers/nnf_node_ec_data_controller.go
NateRoiger: I was trying to preserve nnf-ec's lack of dependency on kubernetes. That way nnf-ec can continue to be an independent program.

I suppose I could pass in the k8sClient/NamespacedName and in nnf-ec implement the CRD Storage Provider... I'm not sure where the CRD would live then; and how would it get deployed?

controllers/nnf_node_ec_data_controller.go
NateRoiger: client & namespaced name

controllers/nnf_node_ec_data_controller.go
roehrich-hpe: A CRD doesn't need to be paired with a reconciler.  There is  an option on kubebuilder/operator-sdk  to say you won't want the reconciler.

NateRoiger: I do think we'll eventually want code in reconciler strictly for delete handling in place of `nnf-ec --purge`. Using purge only deletes namespaces and is therefore problematic.

roehrich-hpe (approved):

PR #43, opened 2022-07-29
RABSW-999: Use WorkflowError in workflow controller

matthew-richerson:
This commit uses the new WorkflowError type in the workflow controller.
This error type does a few things:
- Gives a user facing message about the error
- Specifies whether the error is likely recoverable
- Wraps an implementation level error

If the error is likely unrecoverable, then the workflow status is set to
"error". Otherwise the status is left as "running". In both cases, the
workflow message is set with the user facing message from the WorkflowError.

The message from the implementation level error is placed into a new field,
"error", in the driver status section.


controllers/nnf_workflow_controller.go
```diff
@@ -237,6 +244,7 @@ func (r *NnfWorkflowReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 
 		driverStatus.Reason = "running"
 		driverStatus.Message = ""
+		driverStatus.Error = ""
```
ajfloeder, line 247: Can you use a kubebuilder command to make this the default?

ajfloeder: Withdrawn, this sets each state as it gets to it.

ajfloeder (approved):

NateRoiger (approved):

PR #50, opened 2022-08-26
RABSW-1003: Use ResourceError for NnfNodeStorage and ClientMount

matthew-richerson:
This commit uses the new ResourceError struct from DWS to return extra
error information from the NnfNodeStorage and ClientMount controllers.

This error information is used in the workflow resource. For example, an allocation
request that was too large would result in an error in the workflow:

  drivers:
  ...
  - completed: false
    driverID: nnf
    dwdIndex: 0
    error: 'Could not create StoragePool: Error 406: Not Acceptable, Cause: Insufficient
      capacity available, Internal Error: Insufficient capacity available. Requested:
      0x1c6bf52634000 Available: 0x200000000000'
    lastHB: 0
    message: Insufficient capacity available
    status: Error
    taskID: ""
    watchState: setup
  message: 'DW Directive 0: Insufficient capacity available'
  ready: false
  state: setup
  status: Error


api/v1alpha1/nnf_access_types.go
```diff
@@ -66,8 +67,7 @@ type NnfAccessStatus struct {
 	// Ready signifies whether status.state has been achieved
 	Ready bool `json:"ready"`
 
-	// Error message
-	Message string `json:"message,omitempty"`
+	Error *dwsv1alpha1.ResourceError `json:"error,omitempty"`
```
NateRoiger, line 70: Is a struct considered "empty" if all its members are empty? Wondering if these need to be pointers

matthew-richerson: Not in the "omitempty" sense. If there are fields in the struct that aren't omitempty, then the struct should be a pointer if you want the whole thing to be omitempty.

So in this case, if there is an error, then I want some of the fields in the ResourceError to be required. If there isn't an error, then the whole ResourceError struct should be omitted.

ajfloeder (approved):

roehrich-hpe (approved):

api/v1alpha1/nnf_access_types.go
NateRoiger: OK - wasn't quite sure

NateRoiger (approved):

PR #56, opened 2022-08-30
Revert to Ginkgo V1.

roehrich-hpe:
Kubebuilder does not yet support Ginkgo v2, and sometimes craps out when we
ask it to do new things in this repo when v2 is in use.

Upgrade controller-gen from 0.6.1 to 0.9.2, which causes some annotations to
be updated in the CRDs.


go.mod
```diff
@@ -19,3 +19,84 @@ require (
 	k8s.io/client-go v0.23.6
 	sigs.k8s.io/controller-runtime v0.11.2
 )
+
+require (
```
NateRoiger, line 23: what's this about?

NateRoiger (approved):

go.mod
roehrich-hpe: The flip on line 3 from go 1.16 to 1.17 did that.

ajfloeder (approved):

PR #67, opened 2022-09-12
Updater

NateRoiger:
Use the new dwsutil updater code for all statusUpdaters.


ajfloeder (approved):

matthew-richerson (approved):

controllers/nnf_node_controller.go
```diff
@@ -417,12 +375,8 @@ func updateDrives(node *nnfv1alpha1.NnfNode, statusUpdater *statusUpdater, log l
...
 				}
 
-				if drive.Model != storageController.Model || drive.WearLevel != int64(storageController.NVMeControllerProperties.NVMeSMARTPercentageUsage) {
-					statusUpdater.Update(func(*nnfv1alpha1.NnfNodeStatus) {
-						drive.Model = storageController.Model
-						drive.WearLevel = int64(storageController.NVMeControllerProperties.NVMeSMARTPercentageUsage)
-					})
-				}
+				drive.Model = storageController.Model
+				drive.WearLevel = int64(storageController.NVMeControllerProperties.NVMeSMARTPercentageUsage)
```
bdevcich-hpe, line 379: So the idea is that we're always just setting these (no longer checking if they don't match) and then the `defer CloseWithStatusUpdate` performs the update?

bdevcich-hpe (approved):

controllers/nnf_node_controller.go
NateRoiger: the [updater records the status when it's created](https://github.com/HewlettPackard/dws/blob/8515b1a0003ddc1d5775c0df9c5ae503628c3cac/utils/updater/status_updater.go#L69); then when it's closed it [does a comparison between the old status and the new status](https://github.com/HewlettPackard/dws/blob/8515b1a0003ddc1d5775c0df9c5ae503628c3cac/utils/updater/status_updater.go#L94). Only if the values change does the update actually occur.

bdevcich-hpe: Thank you. I missed the `DeepEqual`. TIL.

Issue #68, opened 2022-09-12
lustre fsname not unique for shared MGS


matthew-richerson:
When using a shared MGS, the Lustre fsname that we use is not unique. The fsname is currently the name specified on the #DW line.

roehrich-hpe:
Ah, never mind that.  It's the name registered with the MGS that matters, not the resource name.

PR #69, opened 2022-09-12
Change fsname for Lustre (#68)

matthew-richerson:
This commit changes the string that is used for the lustre fsname. Instead of using the "name" field in the #DW line, this commit uses the first 8 characters of the UID of the servers resource.

### DIFF
--- a/controllers/nnf_workflow_controller.go
+++ b/controllers/nnf_workflow_controller.go
@@ -719,11 +719,7 @@ func (r *NnfWorkflowReconciler) createNnfStorage(ctx context.Context, workflow *
 				if dwArgs["type"] == "lustre" {
 					nnfAllocSet.NnfStorageLustreSpec.TargetType = strings.ToUpper(s.Spec.AllocationSets[i].Label)
 					nnfAllocSet.NnfStorageLustreSpec.BackFs = "zfs"
-					charsWanted := 8
-					if len(dwArgs["name"]) < charsWanted {
-						charsWanted = len(dwArgs["name"])
-					}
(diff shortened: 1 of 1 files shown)


controllers/nnf_workflow_controller.go
```diff
@@ -719,11 +719,7 @@ func (r *NnfWorkflowReconciler) createNnfStorage(ctx context.Context, workflow *
 				if dwArgs["type"] == "lustre" {
 					nnfAllocSet.NnfStorageLustreSpec.TargetType = strings.ToUpper(s.Spec.AllocationSets[i].Label)
 					nnfAllocSet.NnfStorageLustreSpec.BackFs = "zfs"
-					charsWanted := 8
-					if len(dwArgs["name"]) < charsWanted {
-						charsWanted = len(dwArgs["name"])
-					}
-					nnfAllocSet.NnfStorageLustreSpec.FileSystemName = dwArgs["name"][:charsWanted]
```
NateRoiger, line 726: Should there be a generic "Name" field that still records the `dwsArgs["name"]` value? Just wondering for debug it might be handy. Only asking - I haven't really digested this layer yet to fully comprehend it.

NateRoiger (approved):

roehrich-hpe (approved):

controllers/nnf_workflow_controller.go
matthew-richerson: It wouldn't be used for anything, and we already have a way to tie the nnfStorage back to which #DW it's from with the index in the labels section. That makes me think that it wouldn't be useful, but someone else might think otherwise.

The name is still used on the user side of things for referencing a filesystem with the environment variables. For example: `DW_JOB_coolfs`

bdevcich-hpe (approved):

PR #70, opened 2022-09-15
Dm errors

NateRoiger:
- Ensure all nnf accesses unmount and wait where needed, but don't delete until teardown
- Ensure all nnf dm operations that fail communicate that failure to the workflow
- Adds new NewWorkflowError**f** for simple error formatting like Printf, Sprintf, Errorf, etc.

### DIFF
--- a/controllers/nnf_workflow_controller_helpers.go
+++ b/controllers/nnf_workflow_controller_helpers.go
@@ -1,6 +1,7 @@
 package controllers
 
 import (
+	"context"
 	"fmt"
 	"strconv"
 	"strings"
@@ -9,6 +10,8 @@ import (
 	"github.com/HewlettPackard/dws/utils/dwdparse"
 	nnfv1alpha1 "github.com/NearNodeFlash/nnf-sos/api/v1alpha1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 func handleWorkflowError(err error, driverStatus *dwsv1alpha1.WorkflowDriverStatus) {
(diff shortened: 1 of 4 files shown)


controllers/nnf_workflow_controller_helpers.go
```diff
@@ -118,3 +121,49 @@ func addDirectiveIndexLabel(object metav1.Object, index int) {
...
+func (r *NnfWorkflowReconciler) waitForNnfAccessStateAndReady(ctx context.Context, workflow *dwsv1alpha1.Workflow, index int, state string) (*ctrl.Result, error) {
+
+	accessSuffixes := []string{"-computes"}
+
+	fsType, err := r.getDirectiveFileSystemType(ctx, workflow, index)
+	if err != nil {
+		return nil, nnfv1alpha1.NewWorkflowError("Unable to determine directive file system type").WithError(err)
+	}
+
+	if fsType == "gfs2" || fsType == "lustre" {
```
bdevcich-hpe, line 135: I've noticed this sort of thing in a number of places. Are we going to be supporting additional filesystems at any point? If so, we should keep a list of supported fs so we don't need to do these explicit checks all over the place.

controllers/nnf_workflow_controller_helpers.go
```diff
@@ -118,3 +121,49 @@ func addDirectiveIndexLabel(object metav1.Object, index int) {
...
+		}
+
+		if access.Status.Error != nil {
+			err = fmt.Errorf("Error on NnfAccess %s: %w", client.ObjectKeyFromObject(access).String(), access.Status.Error)
+			return nil, nnfv1alpha1.NewWorkflowError("Could not access file system on nodes").WithError(err)
+		}
+
+		// If we're mounting, we must always wait for ready
+		// If we're unmounting, check that the teardown state label is for this state
+
```
bdevcich-hpe, line 159: remove this line

bdevcich-hpe (approved):

controllers/nnf_workflow_controller_helpers.go
NateRoiger: I've been wanting to change this to a type for the longest time. Then we can do `fsType.IsShared()`

But, that being said - LLNL doesn't require any more file systems from _us_, but they want to use containers to manage _any_ file system i.e. GPFS, UnifyFS

matthew-richerson (approved):

bdevcich-hpe (approved):

ajfloeder (approved):

roehrich-hpe (approved):

ajfloeder (approved):

PR #73, opened 2022-09-15
Added SerialNumber, FirmwareVersion, and Slot to NnfDriveStatus

bdevcich-hpe:
Each new field is sourced from the `StorageController`'s corresponding field.

In the case of `slot`, it is a concatenation of the `StorageController`'s `Location` field. The `SeviceLabel` and `LocationOrdinalValue` are put together. These two fields are sourced from the StorageController's config file - see nnf-ec for more details.

For example, with a label of `J` and ordinal value of `12`, the resulting `slot` is `J12`.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -51,5 +51,17 @@
             },
             "showLog": true
         },
+        {
+            "name": "Debug k8s operator remotely",
+            "type": "go",
+            "request": "attach",
+            "mode": "remote",
+            "remotePath": "",
+            "port": 40000,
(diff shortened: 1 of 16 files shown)


NateRoiger (approved):
Looks good! Just a reminder to update go.mod to point to dws/nnf-ec @ master

ajfloeder (approved):

bdevcich-hpe:
I had to rebase to get the build to work. I think it had to do with bringing in master nnf-ec/dws with @roehrich-hpe's changes to envtest and not having that here as well.

PR #74, opened 2022-09-15
Proposal for adding custom result to workflow start|finish states & dws object deletion code

NateRoiger:
This is a proposal to return additional information for each start|finish state so the logs have more details as to _WHY_ waits and retries are occurring.

(this won't compile as is)

### DIFF
--- a/controllers/nnf_workflow_controller.go
+++ b/controllers/nnf_workflow_controller.go
@@ -61,80 +61,6 @@ const (
 	finalizerNnfWorkflow = "nnf.cray.hpe.com/nnf_workflow"
 )
 
-type result struct {
-	ctrl.Result
-	reason       string
-	object       client.Object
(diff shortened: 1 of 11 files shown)


controllers/nnf_workflow_controller.go
```diff
@@ -344,7 +382,7 @@ func (r *NnfWorkflowReconciler) finishProposalState(ctx context.Context, workflo
 
 	// Wait for the breakdown to be ready
 	if directiveBreakdown.Status.Ready != ConditionTrue {
-		return &ctrl.Result{}, nil
+		return Requeue.withObject(directiveBreakdown).withReason("status pending"), nil
```
bdevcich-hpe, line 385: Could you do something like result.Requeue("status pending") to make this easier?

matthew-richerson (approved):

PR #83, opened 2022-10-11
EC errors from FS creation are fatal, and reconciliation should stop.

roehrich-hpe:

### DIFF
--- a/controllers/nnf_node_storage_controller.go
+++ b/controllers/nnf_node_storage_controller.go
@@ -450,7 +450,7 @@ func (r *NnfNodeStorageReconciler) formatFileSystem(nodeStorage *nnfv1alpha1.Nnf
 		nodeStorage.Status.Error = dwsv1alpha1.NewResourceError("Could not create file system", err).WithFatal()
 		log.Info(nodeStorage.Status.Error.Error())
 
-		return &ctrl.Result{Requeue: true}, nil
+		return &ctrl.Result{}, nil
 	}
 
 	// If the SF ID is empty then we just created the resource. Save the ID in the NnfNodeStorage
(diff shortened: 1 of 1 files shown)


ajfloeder (approved):

bdevcich-hpe (approved):

controllers/nnf_node_storage_controller.go
```diff
@@ -450,7 +450,7 @@ func (r *NnfNodeStorageReconciler) formatFileSystem(nodeStorage *nnfv1alpha1.Nnf
 		nodeStorage.Status.Error = dwsv1alpha1.NewResourceError("Could not create file system", err).WithFatal()
 		log.Info(nodeStorage.Status.Error.Error())
 
-		return &ctrl.Result{Requeue: true}, nil
+		return &ctrl.Result{}, nil
```
matthew-richerson, line 453: Can we do a RequeueAfter instead? This assumes that we can never recover.

roehrich-hpe: If it's an error in the args for zpool, mkfs, mount...then we can't recover.  The workflow has to be torn down, the storageprofile has to be fixed, and then a new workflow has to be created.

But you've put your comment on the place where the EC's data structures are being created.  Do we really expect that if the setup for the filesystem data structures didn't work the first time, it'll work the second time?

matthew-richerson: My comment was meant for both of the places you're taking out the requeue. There are errors other than invalid arguments that could be more transient. For example, a network issue preventing a Lustre OST/MDT mount from contacting the MGS.

matthew-richerson: Sounds good

matthew-richerson (approved):

PR #84, opened 2022-10-13
Add ProgressCollectionInterval to CommandStatus

bdevcich-hpe:
More help for testing data movement configuration/progress.

### DIFF
--- a/api/v1alpha1/nnf_datamovement_types.go
+++ b/api/v1alpha1/nnf_datamovement_types.go
@@ -77,6 +77,10 @@ type NnfDataMovementCommandStatus struct {
 	// The command that was executed during data movement.
 	Command string `json:"command,omitempty"`
 
+	// The interval to wait between progress collection during data movement.
+	// +kubebuilder:default:="5s"
+	ProgressCollectionInterval string `json:"collectInterval,omitempty"`
+
 	// ElapsedTime reflects the elapsed time since the underlying data movement command started.
 	ElapsedTime metav1.Duration `json:"elapsedTime,omitempty"`
 
--- a/config/crd/bases/nnf.cray.hpe.com_nnfdatamovements.yaml
+++ b/config/crd/bases/nnf.cray.hpe.com_nnfdatamovements.yaml
@@ -170,6 +170,11 @@ spec:
                   at a certain frequency to avoid excessive updates to the Data Movement
                   resource.
                 properties:
+                  collectInterval:
+                    default: 5s
+                    description: The interval to wait between progress collection
+                      during data movement.
+                    type: string
                   command:
                     description: The command that was executed during data movement.
                     type: string


api/v1alpha1/nnf_datamovement_types.go
```diff
@@ -77,6 +77,10 @@ type NnfDataMovementCommandStatus struct {
 	// The command that was executed during data movement.
 	Command string `json:"command,omitempty"`
 
+	// The interval to wait between progress collection during data movement.
+	// +kubebuilder:default:="5s"
```
NateRoiger, line 81: Is this just to verify the polling interval? Why isn't that just the configmap.data[whatever-interval-name] value?

NateRoiger: I don't think default's can apply to Status subresources

bdevcich-hpe: Yes, this is to verify that the value out of the config map was read/parsed correctly. Otherwise, the only other way I can think of is to verify that the `Updating progress ...` output is in the log and that the timestamps between those log entries are equal to the interval set.

bdevcich-hpe: > I don't think default can apply to Status subresources

I'll get rid of this.

NateRoiger: > Yes, this is to verify that the value out of the config map was read/parsed correctly. 

OK - not sure it's really necessary beyond debug, but it doesn't hurt.

NateRoiger (approved):

matthew-richerson (approved):

api/v1alpha1/nnf_datamovement_types.go
bdevcich-hpe: I’m not a huge fan either but I have a test case for this and I don’t know how else to verify the value.

NateRoiger: For _invalid_ options we'd need a config map validator - which doesn't exist. We could use a custom resource & webhook, but that seems overkill if you ask me - especially since these values are _our_ knobs and not WLM/User knobs.

For _valid_ options can we can just assume it's working? We don't really gain anything in seeing 5s -> 25s, when just reading the config would say 25s. Right?

I wonder if instead of validating the config map, we just prevent data movement until the config is valid. The DM status could be "Invalid Config" or something until it's corrected.

bdevcich-hpe: If anything in the config map is invalid, the behavior I have coded up now is to use the default value. So if you were to supply "25aaa", it would just default to "5s" since it couldn't properly parse that and turn it into a duration.

I could change it to just bomb out with an error if anything is invalid. That's probably more simple.

PR #85, opened 2022-10-17
Allow NnfStorageProfile to pass args for mkfs, lvm, lustre

roehrich-hpe:
The NnfStorageProfile has been changed to take whole commandlines for volume and filesystem creation.

The placeholder_nnfstorageprofile.yaml default profile has been updated with commandlines which create the basic volumes and filesystems that EC was creating before this switch to passing in whole commandlines. This profile continues to be the default profile that is installed by nnf-deploy.

The NnfStorage and NnfNodeStorage resources have labels that name the pinned profile that is being used with the workflow.

The nnf_node_storage_controller chooses appropriate commandlines from the profile when constructing the OEM object for EC.


NateRoiger (approved):

bdevcich-hpe (approved):

api/v1alpha1/nnfstorageprofile_types.go
```diff
@@ -24,6 +24,22 @@ import (
...
+	// "mirror", or "draid".  See zpoolconcepts(7).
+	ZpoolCreate string `json:"zpoolCreate,omitempty"`
+
+	// Mkfs specifies the mkfs.lustre commandline, minus the "mkfs.lustre".
+	// Use the --mkfsoptions argument to specify the zfs create options.  See zfsprops(7).
+	Mkfs string `json:"mkfs,omitempty"`
+
+	// Mount specifies mount options for the mount-utils library.
+	// Use one array element per option. Do not prepend the options with "-o".
+	Mount []string `json:"mount,omitempty"`
```
matthew-richerson, line 40: I think it would be more user friendly if these were all of the same format. Either a string of all the arguments or an array of each argument.

matthew-richerson:
I'm not seeing how the mount options make it out to the compute nodes. Is that in a different PR, or am I missing something?

roehrich-hpe:
> I'm not seeing how the mount options make it out to the compute nodes. Is that in a different PR, or am I missing something?

I think we don't have that.

NateRoiger (approved):

bdevcich-hpe (approved):

matthew-richerson (approved):

PR #86, opened 2022-10-20
RABSW-902: Add references for PersistentStorage consumers

matthew-richerson:
This commit adds an object reference to the PersistentStorage resource for each "persistentdw" directive. The "destroy_persistent" directive will not complete until the consumer reference list in the PersistentStorage is empty.


controllers/integration_test.go
```diff
@@ -146,6 +146,23 @@ var _ = Describe("Integration Test", func() {
...
+				Name:      storageName,
+				Namespace: w.Namespace,
+			},
+		}
+
+		By(fmt.Sprintf("Retrieving PSI '%s'", storageName))
+		Expect(k8sClient.Get(context.TODO(), client.ObjectKeyFromObject(persistentInstance), persistentInstance)).To(Succeed(), "PersistentStorageInstance created")
+
+		By("Checking PSI has the correct consumer reference")
+		Expect(len(persistentInstance.Spec.ConsumerReferences)).To(Equal(1))
```
NateRoiger, line 161: There's a `HaveLen()` matcher for checking arrays
`Expect(persistentInstance.Spec.ConsumerReferences).To(HaveLen(1))`

controllers/nnf_persistentstorageinstance_controller_test.go
```diff
@@ -0,0 +1,125 @@
+package controllers
```
NateRoiger, line 1: CONTROLLER UNIT TEST! YES!

Does Copyright belong in _test.go?

controllers/integration_test.go
matthew-richerson: Nice. I'll use that.

controllers/nnf_workflow_controller_helpers.go
```diff
@@ -699,3 +699,86 @@ func (r *NnfWorkflowReconciler) waitForNnfAccessStateAndReady(ctx context.Contex
...
+	for _, existingReference := range persistentStorage.Spec.ConsumerReferences {
+		if existingReference == reference {
+			found = true
+		}
+	}
+
+	if !found {
+		persistentStorage.Spec.ConsumerReferences = append(persistentStorage.Spec.ConsumerReferences, reference)
+
+		err := r.Update(ctx, persistentStorage)
```
NateRoiger, line 731: Simply `return r.Update(ctx, ps)` ?

controllers/nnf_persistentstorageinstance_controller_test.go
matthew-richerson: Yep, I'll add it.

controllers/nnf_workflow_controller_helpers.go
```diff
@@ -699,3 +699,86 @@ func (r *NnfWorkflowReconciler) waitForNnfAccessStateAndReady(ctx context.Contex
...
+	reference := corev1.ObjectReference{
+		Name:      indexedResourceName(workflow, index),
+		Namespace: workflow.Namespace,
+	}
+
+	found := false
+	for i, existingReference := range persistentStorage.Spec.ConsumerReferences {
+		if existingReference == reference {
+			persistentStorage.Spec.ConsumerReferences = append(persistentStorage.Spec.ConsumerReferences[:i], persistentStorage.Spec.ConsumerReferences[i+1:]...)
+			found = true
```
NateRoiger, line 758: Can't there only ever be one existingReference? So you could do the remove & update here, without the `found` flag?

NateRoiger: I guess it reflects the `addPersistent...` code. So maybe leave it as is.

NateRoiger (approved):

controllers/nnf_workflow_controller_helpers.go
matthew-richerson: Yeah, only one of the references should match, so it would be safe to modify it in the for loop and return. I'm not partial to either way of doing it.

roehrich-hpe (approved):

PR #87, opened 2022-10-21
Allow a colocation constraint to be specified for an MDT.

roehrich-hpe:
Add a field named "exclusiveMdt" to the storage profile to indicate that the MDT must be marked in the directive breakdown as being the only lustre target on its server.

This is similar to the way an MGT is marked in the directive breakdown.

### DIFF
--- a/api/v1alpha1/nnfstorageprofile_webhook.go
+++ b/api/v1alpha1/nnfstorageprofile_webhook.go
@@ -101,9 +101,6 @@ func (r *NnfStorageProfile) validateContentLustre() error {
 	if r.Data.LustreStorage.CombinedMGTMDT && len(r.Data.LustreStorage.ExternalMGS) > 0 {
 		return fmt.Errorf("cannot set both combinedMgtMdt and externalMgs")
 	}
-	if r.Data.LustreStorage.CombinedMGTMDT && r.Data.LustreStorage.ExclusiveMDT {
-		return fmt.Errorf("cannot set both combinedMgtMdt and exclusiveMdt; the target will already be constrained because it is an MGT")
-	}
 
 	return nil
 }
--- a/api/v1alpha1/nnfstorageprofile_webhook.go
+++ b/api/v1alpha1/nnfstorageprofile_webhook.go
@@ -64,7 +64,7 @@ func (r *NnfStorageProfile) ValidateUpdate(old runtime.Object) error {
 		// ownerReferences, and labels, but do not allow Data to be
 		// updated.
 		if !reflect.DeepEqual(r.Data, obj.Data) {
-			msg := "Update on pinned resource not allowed"
+			msg := "update on pinned resource not allowed"
 			err := fmt.Errorf(msg)
 			nnfstorageprofilelog.Error(err, "invalid")
 			return err
@@ -89,7 +89,7 @@ func (r *NnfStorageProfile) ValidateDelete() error {
 func (r *NnfStorageProfile) validateContent() error {
 
 	if r.Data.Default && r.Data.Pinned {
-		return fmt.Errorf("The NnfStorageProfile cannot be both default and pinned.")
+		return fmt.Errorf("the NnfStorageProfile cannot be both default and pinned")
 	}
 	if err := r.validateContentLustre(); err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


ajfloeder:
Do we need to add this to #DW for both `jobdw` and `persistentdw` for completeness so the user can specify it in their #DW without adding it to a profile? Similar to `combinedMGTMDT`?

roehrich-hpe:
> Do we need to add this to #DW for both `jobdw` and `persistentdw` for completeness so the user can specify it in their #DW without adding it to a profile? Similar to `combinedMGTMDT`?

I was thinking the opposite.  Remove the "combined_mgtmdt" and "external_mgs" from the #DW line and make them strictly  configurable via profile.

On the other hand, so far the only use-case mentioned to me for this exclusive-mdt option is for Cray's own benchmarkers.  So I wouldn't put it in the #DW for that.

ajfloeder:
> 

I like that. It constrains the workflow creators to the configured profiles and pushes the support/test burden on the profile process.

controllers/directivebreakdown_controller.go
```diff
@@ -421,12 +421,17 @@ func (r *DirectiveBreakdownReconciler) populateStorageBreakdown(ctx context.Cont
 		var lustreComponents []lustreComponentType
 		lustreComponents = append(lustreComponents, lustreComponentType{"AllocateAcrossServers", breakdownCapacity, "ost", ""})
 
+		mdtColocationKey := ""
+		if nnfStorageProfile.Data.LustreStorage.ExclusiveMDT {
+			mdtColocationKey = "lustre-mdt"
```
matthew-richerson, line 426: The current use of our exclusive colocation constraint for an allocation set looks like this:
```
colocation:
- key: lustre-mgt
  type: exclusive
```

That means that this allocation set can't be colocated with any other allocation sets that have the same exclusive colocation constraint key. Practically, that allows the MGTs to exist on different nodes, but lets them be shared with OSTs and MDTs both from this filesystem and others. What you have here would do the same for MDTs. Two MDTs couldn't be on the same node, but an MDT could be colocated with one or more OSTs. I don't think that's what you want.

If you wanted the MDT to not colocate with an OST or MGT, then each of those allocation sets would need an exclusive colocation constraint that shared the same key. However, that would mean that the OSTs would now be restricted to only one OST per server. That might not be what the performance team wants.

Also, the exclusive colocation constraints would only apply for Lustre targets from that single file system. If another Lustre filesystem was allocated without the exclusive_mdt option, then it wouldn't have the colocation constraints in the directive breakdown, and it could place Lustre targets on the servers with the MDTs from the first file system. Maybe that's desirable, maybe not? The performance testers will probably only make a single file system at a time, so I don't think it matters for their use case.

I think a better option would be to add an exclusive colocation constraint without a key. That would mean that the allocation set was exclusive with everything, not just scoped to allocation sets with the same key. The Flux team would have to be told how to interpret that. If it was valuable, we could have a "within" field that scoped it to either the current filesystem or to all file systems. Doing it this way would solve the problem of having the MDTs be exclusive while still allowing more than one OST on a Rabbit.

```
    allocationSets:
    - allocationStrategy: AllocateAcrossServers
      constraints:
        labels:
        - dws.cray.hpe.com/storage=Rabbit
      label: ost
      minimumCapacity: 50000000000
    - allocationStrategy: AllocateSingleServer
      constraints:
        colocation:
        - type: exclusive
          within: directive
        labels:
        - dws.cray.hpe.com/storage=Rabbit
      label: mdt
      minimumCapacity: 5368709120
    - allocationStrategy: AllocateSingleServer
      constraints:
        colocation:
        - key: lustre-mgt
          type: exclusive
        labels:
        - dws.cray.hpe.com/storage=Rabbit
      label: mgt
      minimumCapacity: 1073741824
```

bdevcich-hpe (approved):

roehrich-hpe:
That last commit is just a rebase with master.

ajfloeder (approved):

bdevcich-hpe (approved):

api/v1alpha1/nnfstorageprofile_webhook.go
```diff
@@ -99,7 +99,10 @@ func (r *NnfStorageProfile) validateContent() error {
 
 func (r *NnfStorageProfile) validateContentLustre() error {
 	if r.Data.LustreStorage.CombinedMGTMDT && len(r.Data.LustreStorage.ExternalMGS) > 0 {
-		return fmt.Errorf("Cannot set both combinedMgtMdt and externalMgs")
+		return fmt.Errorf("cannot set both combinedMgtMdt and externalMgs")
+	}
+	if r.Data.LustreStorage.CombinedMGTMDT && r.Data.LustreStorage.ExclusiveMDT {
+		return fmt.Errorf("cannot set both combinedMgtMdt and exclusiveMdt; the target will already be constrained because it is an MGT")
```
matthew-richerson, line 105: I don't think this is true. The MGT will be prevented from colocating with other MGTs, but it could shared a node with an OST.

matthew-richerson (approved):

api/v1alpha1/nnfstorageprofile_webhook.go
roehrich-hpe: Sure.  I'll remove that constraint.

PR #96, opened 2022-11-10
RABSW-1086: Remove combined_mgtmdt and external_mgs directive arguments

matthew-richerson:
Combined_mgtmdt and external_mgs can be specified in the storage profile. Removing these options from the directive allows the admins to have greater control of what end users can and cannot do.

Signed-off-by: Matt Richerson <matthew.richerson@dhcp-10-113-10-52.its.hpecorp.net>


bdevcich-hpe (approved):

NateThornton (approved):
-106 lines, hell yeah

PR #100, opened 2022-12-02
RABSW-1083: Change group and owner of storage

matthew-richerson:
This commit passes the group ID and owner ID fields down to the NnfNodeStorage so it can be used to set the owner/group of a file system. For XFS and gfs2, this is done through nnf-ec. For Lustre file systems, all Lustre targets have to complete before the owner/group can be set. The Lustre client mounts the file system on the Rabbit node hosting OST 0, and the owner/group are set from there. This is coordinated through a "SetOwnerGroup" flag in the NnfNodeStorage that is set once all the Lustre targets are ready.

### DIFF
--- a/api/v1alpha1/nnf_node_storage_types.go
+++ b/api/v1alpha1/nnf_node_storage_types.go
@@ -43,6 +43,8 @@ type NnfNodeStorageSpec struct {
 	// Group ID for file system
 	GroupID uint32 `json:"groupID"`
 
+	// Set the owner and group permissions specified by UserID and GroupID. This is for
+	// Lustre file systems only, and should be set only after all Lustre targets are created.
 	// +kubebuilder:default:=false
 	SetOwnerGroup bool `json:"setOwnerGroup"`
 
--- a/api/v1alpha1/nnf_node_storage_types.go
+++ b/api/v1alpha1/nnf_node_storage_types.go
@@ -37,6 +37,15 @@ type NnfNodeStorageSpec struct {
 	// +kubebuilder:validation:Minimum:=0
 	Count int `json:"count"`
 
+	// User ID for file system
+	UserID uint32 `json:"userID"`
+
+	// Group ID for file system
+	GroupID uint32 `json:"groupID"`
+
+	// +kubebuilder:default:=false
+	SetOwnerGroup bool `json:"setOwnerGroup"`
+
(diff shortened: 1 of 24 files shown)


api/v1alpha1/nnf_node_storage_types.go
```diff
@@ -37,6 +37,15 @@ type NnfNodeStorageSpec struct {
...
 	Count int `json:"count"`
 
+	// User ID for file system
+	UserID uint32 `json:"userID"`
+
+	// Group ID for file system
+	GroupID uint32 `json:"groupID"`
+
+	// +kubebuilder:default:=false
+	SetOwnerGroup bool `json:"setOwnerGroup"`
```
NateThornton, line 47: +Description

ajfloeder, line 47: Why wouldn't we always set the owner/group? Thus, you wouldn't need this.

NateThornton (approved):

api/v1alpha1/nnf_node_storage_types.go
matthew-richerson: It's only for Lustre file systems, and it only gets set for one of the nnfNodeStorage resources. The Lustre file system has lots of different Lustre targets (MDTs and OSTs) each with an nnfNodeStorage, but we only need to have one client to mount the file system and set the permissions. OST 0 is always picked as the nnfNodeStorage to make "SetOwnerGroup".

controllers/nnf_node_storage_controller.go
```diff
@@ -584,6 +601,49 @@ func (r *NnfNodeStorageReconciler) formatFileSystem(ctx context.Context, nodeSto
...
+
+	source := nodeStorage.Spec.LustreStorage.MgsNode + ":/" + nodeStorage.Spec.LustreStorage.FileSystemName
+
+	if !mounted {
+		if err := mounter.Mount(source, target, "lustre", nil); err != nil {
+			log.Error(err, "Mount failed")
+			return err
+		}
+	}
+	defer func() { err = mounter.Unmount(target) }()
```
roehrich-hpe, line 637: You don't have 'err' as a named return value, and if you did it would overwrite an error from os.Chown below.

matthew-richerson: Oops, I forgot to add the fix here. I'll fix it.

PR #104, opened 2022-12-09
Fix: XFS doesn't unmount after data-in finishes

NateThornton:
The offending code block is both wrong and not needed as the _unmount-when-necessary_ logic always uses the `nnfv1alpha1.DataMovementTeardownStateLabel`


ajfloeder (approved):

roehrich-hpe:
@NateThornton Beginning with this commit, I'm getting a consistent test failure here:
```
  Timed out after 20.001s.
  Waiting on ready status state PostRun
  Expected
      <bool>: false
  to be true
  In [It] at: /Users/droehrich/work/rabsw/nnf-sos-2.git/controllers/integration_test.go:765
```

PR #105, opened 2022-12-09
RABSW-1095: Unlock NnfStorage when NnfAccess is unmounted

matthew-richerson:
The NnfStorage is locked for XFS file systems when an NnfAccess has the file system mounted. This prevents the file system from being mounted in two places at once which could lead to corruption issues. The lock was only removed when the NnfAccess was deleted, but it needs to be removed when the NnfAccess is unmounted.

### DIFF
--- a/controllers/nnf_access_controller.go
+++ b/controllers/nnf_access_controller.go
@@ -217,6 +217,13 @@ func (r *NnfAccessReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		log.Info("State achieved", "State", access.Status.State)
 	}
 
+	if access.Status.State == "unmounted" {
+		// Unlock the NnfStorage so it can be used by another NnfAccess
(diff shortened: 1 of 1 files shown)


ajfloeder (approved):

controllers/nnf_access_controller.go
```diff
@@ -301,11 +308,15 @@ func (r *NnfAccessReconciler) unlockStorage(ctx context.Context, access *nnfv1al
 		if annotations == nil {
 			annotations = make(map[string]string)
 		}
-		_, exists := annotations[NnfAccessAnnotation]
+		value, exists := annotations[NnfAccessAnnotation]
 		if !exists {
 			return nil
 		}
 
+		if value != access.Name+"/"+access.Namespace {
```
NateThornton, line 316: Can you explain this?

bdevcich-hpe (approved):

NateThornton (approved):

PR #106, opened 2022-12-12
Fix for the NNF Node Status reporting 'Ready' when not ready

NateThornton:
- Ensure the NNF Node `status.status` field is "Starting" at start of day
- Ensure the NNF Node `status.status` and `status.health` fields reflective of the current NNF Storage Service values while it is initializing
- Clean up some rookie error handling mistakes in the NNF Node `Start()` routine
- Add some default conflict retry in updating the NNF Node in the `Start()` routine. I have never observed this but it's likely safer than the current implementation
- Add the associated Pod name to NNF Node's default printcolumns

### DIFF
--- a/config/crd/bases/nnf.cray.hpe.com_nnfnodes.yaml
+++ b/config/crd/bases/nnf.cray.hpe.com_nnfnodes.yaml
@@ -28,13 +28,14 @@ spec:
       jsonPath: .status.status
       name: STATUS
       type: string
-    - description: Parent pod name
-      jsonPath: .spec.pod
-      name: POD
-      type: string
     - jsonPath: .metadata.creationTimestamp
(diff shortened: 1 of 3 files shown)


matthew-richerson (approved):

config/crd/bases/nnf.cray.hpe.com_nnfnodes.yaml
```diff
@@ -28,6 +28,10 @@ spec:
       jsonPath: .status.status
       name: STATUS
       type: string
+    - description: Parent pod name
```
roehrich-hpe, line 31: make manifests, to pick up the priority change.

roehrich-hpe (approved):

PR #119, opened 2023-01-11
RABSW-1107: Add container directive validation

bdevcich-hpe:
These changes allow a user to create a container workflow and have it go ready in the `Proposal` state. It validates that:
- the supplied container profile exists
- if any non-optional storage in the container profile exists, ensure it is present in the supplied arguments to the container directive
- any supplied storage argument in the container directive must exist in the container profile

In order to support validation of the `container` directive, the `NnfStorageProfile` pinning and default functionality has been duplicated and refactored for `NnfContainerProfile`. Ideally in the future, we combine the two.

The following directives were used to test this container workflow:
```
#DW jobdw name=blake-gfs2 type=gfs2 capacity=50GB
#DW persistentdw name=blake-persistent
#DW container name=blake-container profile=nnfcontainerprofile-sample \
      JOB_DW_foo-local-storage=blake-gfs2 \
      PERSISTENT_DW_foo-persistent-storage=blake-persistent
```

This requires that a persistent storage instance is present prior to creating the container workflow. This means that you need a workflow with a `create_persistent` directive prior to creating a container.

Work still needs to be done on the directive breakdown controller.

### DIFF
--- a/api/v1alpha1/nnfcontainerprofile_types.go
+++ b/api/v1alpha1/nnfcontainerprofile_types.go
@@ -26,6 +26,14 @@ import (
 
 // NnfContainerProfileSpec defines the desired state of NnfContainerProfile
 type NnfContainerProfileData struct {
+	// Default is true if this instance is the default resource to use
+	// +kubebuilder:default:=false
+	Default bool `json:"default,omitempty"`
+
+	// Pinned is true if this instance is an immutable copy
+	// +kubebuilder:default:=false
+	Pinned bool `json:"pinned,omitempty"`
+
 	// List of possible filesystems supported by this container profile
 	Storages []NnfContainerProfileStorage `json:"storages,omitempty"`
 
--- a/api/v1alpha1/nnfcontainerprofile_types.go
+++ b/api/v1alpha1/nnfcontainerprofile_types.go
@@ -26,10 +26,6 @@ import (
 
 // NnfContainerProfileSpec defines the desired state of NnfContainerProfile
 type NnfContainerProfileData struct {
-	// Default is true if this instance is the default resource to use
-	// +kubebuilder:default:=false
-	Default bool `json:"default,omitempty"`
-
 	// Pinned is true if this instance is an immutable copy
 	// +kubebuilder:default:=false
 	Pinned bool `json:"pinned,omitempty"`
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -113,6 +113,8 @@ spec:
             value: "2000"
           - name: NNF_STORAGE_PROFILE_NAMESPACE
             value: nnf-system
+          - name: NNF_CONTAINER_PROFILE_NAMESPACE
+            value: nnf-system
         ports:
           - containerPort: 50057
             name: nnf-ec
--- a/controllers/nnf_workflow_controller_helpers.go
+++ b/controllers/nnf_workflow_controller_helpers.go
@@ -204,11 +204,11 @@ func (r *NnfWorkflowReconciler) validateContainerDirective(ctx context.Context, 
 	// Ensure the supplied profile exists or use the default
 	profile, err := findContainerProfileToUse(ctx, r.Client, args)
 	if err != nil {
-		return nnfv1alpha1.NewWorkflowError(err.Error())
+		return nnfv1alpha1.NewWorkflowError(err.Error()).WithFatal()
 	}
 
-	// Check to see if the directive argument is in the list of storages in the container profile
-	checkDirectiveIsInProfile := func(storageName string) error {
+	// Check to see if the container storage argument is in the list of storages in the container profile
+	checkStorageIsInProfile := func(storageName string) error {
 		for _, storage := range profile.Data.Storages {
 			if storage.Name == storageName {
 				return nil
@@ -217,9 +217,10 @@ func (r *NnfWorkflowReconciler) validateContainerDirective(ctx context.Context, 
 		return fmt.Errorf("storage '%s' not found in container profile '%s'", storageName, profile.Name)
 	}
 
-	// Find the wildcard JOB_DW and PERSISTENT_DW arguments and verify that JOB_DW and PERSISTENT_DW
-	// instances exist and that they are found in the container profile
-	var directiveStorages []string // use this list later to validate non-optional storages
+	// Find the wildcard JOB_DW and PERSISTENT_DW arguments. Verify that they exist in the
+	// preceeding directives and in the container profile. For persistent arguments, ensure there
+	// is a persistent storage instance.
+	var suppliedStorageArguments []string // use this list later to validate non-optional storages
 	for arg, storageName := range args {
 		switch arg {
 		case "command", "name", "profile":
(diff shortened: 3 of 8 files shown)


api/v1alpha1/nnfcontainerprofile_types.go
```diff
@@ -26,6 +26,14 @@ import (
 
 // NnfContainerProfileSpec defines the desired state of NnfContainerProfile
 type NnfContainerProfileData struct {
+	// Default is true if this instance is the default resource to use
+	// +kubebuilder:default:=false
+	Default bool `json:"default,omitempty"`
```
matthew-richerson, line 31: does default make sense for containers?

controllers/nnf_workflow_controller_helpers.go
```diff
@@ -198,36 +198,81 @@ func (r *NnfWorkflowReconciler) validateStagingDirective(ctx context.Context, wf
...
+	profile, err := findContainerProfileToUse(ctx, r.Client, args)
+	if err != nil {
+		return nnfv1alpha1.NewWorkflowError(err.Error())
 	}
 
-	// Find the wildcard JOB_DW and PERSISTENT_DW arguments
-	for arg, val := range args {
-		// log.Log.Info("BLAKE: arg", "arg", arg)
+	// Check to see if the directive argument is in the list of storages in the container profile
+	checkDirectiveIsInProfile := func(storageName string) error {
```
matthew-richerson, line 211: "Directive" here threw me off. I think of a directive as the whole #DW line. This seems to be checking whether a storage allocation is included in the profile.

controllers/nnf_workflow_controller_helpers.go
```diff
@@ -198,36 +198,81 @@ func (r *NnfWorkflowReconciler) validateStagingDirective(ctx context.Context, wf
...
+	// Ensure that any storage in the profile that isn't marked as optional is present in the directives
+	checkNonOptionalStorages := func(directiveStorageNames []string) error {
+		for _, storage := range profile.Data.Storages {
+			if !storage.Optional {
+				foundInDirectives := false
+
+				for _, directive := range directiveStorageNames {
+					if storage.Name == directive {
+						foundInDirectives = true
+						break
```
matthew-richerson, line 260: You could make this a little simpler by just returning here. Then you wouldn't need the "foundInDirectives" variable.

matthew-richerson (approved):

api/v1alpha1/nnfcontainerprofile_types.go
bdevcich-hpe: I thought that a default profile that contains only (or mostly) optional storages might be useful. But perhaps not.

controllers/nnf_workflow_controller_helpers.go
bdevcich-hpe: Good point. I will change to something like `checkStorageIsInProfile`.

controllers/nnf_workflow_controller_helpers.go
bdevcich-hpe: A return here would break out of the outer loop as well, which I don't think we want here. We need to make sure that any non-optional storage is included as an argument in the container directive. Alternatively, we could use a break label.

Or am I missing something here?

matthew-richerson: Nope. You're right. Disregard

config/manager/manager.yaml
```diff
@@ -113,6 +113,8 @@ spec:
             value: "2000"
           - name: NNF_STORAGE_PROFILE_NAMESPACE
             value: nnf-system
+          - name: NNF_CONTAINER_PROFILE_NAMESPACE
```
NateThornton, line 116: If this is a requirement we should make sure it's in the RFC (description and sample). Maybe even make validating webhook?

@roehrich-hpe How did you handle this requirement?

controllers/nnf_workflow_controller_helpers.go
```diff
@@ -198,36 +198,81 @@ func (r *NnfWorkflowReconciler) validateStagingDirective(ctx context.Context, wf
...
-					return nnfv1alpha1.NewWorkflowError(fmt.Sprintf("persistentdw directive mentioning '%s' not found", name)).WithFatal()
+				if findDirectiveIndexByName(wf, storageName) == -1 {
+					return nnfv1alpha1.NewWorkflowError(fmt.Sprintf("persistentdw directive mentioning '%s' not found", storageName)).WithFatal()
 				}
+				if err := checkDirectiveIsInProfile(arg); err != nil {
+					return nnfv1alpha1.NewWorkflowError(err.Error()).WithFatal()
+				}
+				directiveStorages = append(directiveStorages, arg)
 			}
 			// TODO: just ignore anything that doesn't start with JOB_DW/PERSISTENT_DW?
```
NateThornton, line 247: WithFatal() for now so you don't escape this trap

config/manager/manager.yaml
bdevcich-hpe: I do have plans to add a webhook in another PR.

api/v1alpha1/nnfcontainerprofile_types.go
bdevcich-hpe: I'll remove this. We got this for free with @roehrich-hpe's storage profile implementation, so I figured why not.

controllers/nnf_workflow_controller_helpers.go
NateThornton: (this is totally developer preference, so ignore me if you like)

These nested loops with the `found := false` pattern is why I prefer using functions

```go
findInDirectives := func(name string) bool {
  for _, directive := range directiveStorageNames {
    if name == directive {
      return true
    }
    
    return false
}

for _, storage := range profile.Data.Storages {
  if !storage.Optional {    
    if !findInDirectives(storage.name) {
      return fmt.Errorf(...)
    }
  }
}
```

controllers/nnfcontainerprofile_helpers.go
```diff
@@ -0,0 +1,169 @@
...
+		return nil, err
+	}
+	if !NnfContainerProfile.Data.Pinned {
+		return nil, fmt.Errorf("Expected pinned NnfContainerProfile, but it was not pinned: %s", pinnedName)
+	}
+	return NnfContainerProfile, nil
+}
+
+// createPinnedProfile finds the specified profile and makes a pinned copy of it.
+func createPinnedContainerProfile(ctx context.Context, clnt client.Client, clntScheme *runtime.Scheme, args map[string]string, owner metav1.Object, pinnedName string) (*nnfv1alpha1.NnfContainerProfile, error) {
```
NateThornton, line 98: What, no generics?

config/manager/manager.yaml
roehrich-hpe: I have a webhook for nnfstorageprofile, but it doesn't validate that the NNF_STORAGE_PROFILE_NAMESPACE is set.
I suppose it could do that.

Blake, it's possible that your validation steps can be expressed with CEL and that you wouldn't need a webhook.  See things like:
https://kubernetes.io/blog/2022/09/29/enforce-immutability-using-cel/
https://blog.rewanthtammana.com/kubernetes-crd-validation-with-cel-and-kubebuilder-marker-comments

That would be better than a webhook, if CEL can express everything you require.

controllers/nnfcontainerprofile_helpers.go
roehrich-hpe: oh.   Go for it, Nate.

controllers/nnf_workflow_controller_helpers.go
bdevcich-hpe: I like this. I'll go that route.

config/manager/manager.yaml
bdevcich-hpe: Thanks for those pointers, @roehrich-hpe. I've made a ticket for webhook/CEL.

controllers/nnfcontainerprofile_helpers.go
bdevcich-hpe: Eventually, yes :)

I did put a TODO in the top of the file. I don't like having to duplicate all this, either.

NateThornton:
```go
type NnfContainerProfile struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`
}
​
func (in *NnfContainerProfile) DeepCopyInto(out *NnfContainerProfile) {}
func (in *NnfContainerProfile) DeepCopy() *NnfContainerProfile        { return nil }
func (in *NnfContainerProfile) DeepCopyObject() runtime.Object        { return nil }
​
type Profile interface {
	client.Object
	nnfv1alpha1.NnfStorageProfile | NnfContainerProfile
}
​
func findProfileToUseGeneric[P Profile](ctx context.Context, c client.Client) (*P, error) {
​
	var profile P
​
	if err := c.Get(ctx, types.NamespacedName{Name: "test", Namespace: "default"}, profile); err != nil {
		return nil, err
	}
​
	return &profile, nil
}
```

roehrich-hpe: Those deepcopy's should already be in zz_generated.deepcopy.go.

NateThornton (approved):

PR #126, opened 2023-01-18
Ensure the NNF Node resource fencing status is cleared prior updating…

NateThornton:
- Ensure the NNF Node resource fencing status is cleared prior to updating the Storage resource.
- Don't use a status updater for the NNF Node; do the status update manually. While legal, it's unconventional to have two of these in the same reconciler.


controllers/nnf_node_slc.go
```diff
@@ -156,18 +152,31 @@ func (r *NnfNodeSLCReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
-			nnfNode.Status.Fenced = false
 
-			storage.Status.RebootRequired = false
-			storage.Status.Message = ""
+			if nnfNode.Status.Fenced {
+				log.Info("Clearing fencing status")
+
+				nnfNode.Status.Fenced = false
+
+				if err := r.Status().Update(ctx, nnfNode); err != nil {
```
matthew-richerson, line 164: We talked about this over slack, but I'm having second thoughts now. In our other reconcilers, the defer status updater is for the resource that is being reconciled. That's the NnfNode in this case. The Storage resource is the weird case where we have a defer status updater, but it's not the resource being reconciled.

matthew-richerson (approved):
Looks good other than my comment. I'll leave it up to you on how you want to handle it. Ideally there would be a separate Storage resource reconciler that handled updating the Storage resource status section based on the spec. That might not be possible here.

bdevcich-hpe (approved):

matthew-richerson (approved):

PR #129, opened 2023-01-19
DRAFT DRAFT: Do not review

matthew-richerson:


NateThornton:
How did we miss this one? Are we missing a unit test somewhere?

matthew-richerson:
> How did we miss this one? Are we missing a unit test somewhere?

It's been on my list of things to do, it just wasn't as important as owner/group for xfs/gfs2 and lustre. There are now 3 different ways that owner/group are set. xfs/gfs2 set it down in nnf-ec when the mkfs is done. Lustre sets it from a temporary lustre client after all the lustre targets are set up. And now raw allocations are done on the compute node by Clientmountd since the chown() will only apply to the current OS.

There's still more work to do too. I have to set the owner/group on the Rabbit node for user containers.

NateThornton (approved):

bdevcich-hpe (approved):

PR #133, opened 2023-01-24
Handle Retryable EC errors for File Share

NateThornton:
I'm thinking this is just a Part 1 - I really should be adding retryable error handling to all nnf-ec objects handling. Which means a function would be handy.

But this also introduces a problem - what if the Retryable occurs for ever (as currently happening on rabbit-node-1); Do we need a Retry Limit? Would it be time based, count based?

### DIFF
--- a/controllers/nnf_node_storage_controller.go
+++ b/controllers/nnf_node_storage_controller.go
@@ -945,7 +945,7 @@ func (r *NnfNodeStorageReconciler) handleCreateError(storage *nnfv1alpha1.NnfNod
 	// If this is really Fatal, we should not retry. But not all of nnf-ec is supports the
 	// retryable classification of errors. Instead we mark the error as Fatal() but continue
 	// to retry with a modest delay. If the resource creation error occurs perpetually, an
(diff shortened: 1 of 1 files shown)


ajfloeder (approved):

controllers/nnf_node_storage_controller.go
```diff
@@ -902,6 +892,30 @@ func (r *NnfNodeStorageReconciler) getFileSystem(ss nnf.StorageServiceApi, id st
...
+	}()
+
+	controllerError := &ec.ControllerError{}
+	if errors.As(err, &controllerError) && controllerError.IsRetryable() {
+		return &ctrl.Result{RequeueAfter: controllerError.RetryDelay()}, nil
+	}
+
+	resourceError = resourceError.WithFatal()
+
+	// If this is really Fatal, we should not retry. But not all of nnf-ec is supports the
```
ajfloeder, line 910:
```suggestion
	// If this is really Fatal, we should not retry. But not all of nnf-ec supports the
```

matthew-richerson (approved):

ajfloeder (approved):

PR #136, opened 2023-01-30
RABSW-1124: Change NnfAccess TeardownState for servers

matthew-richerson:
The data movement code was mounting and unmounting the Rabbit nodes during the DataIn and DataOut phases of the workflow. A stale workflow resource in the client cache could cause the NnfAccess to be re-mounted after it had already been unmounted. This commit changes the NnfAccess Teardown state logic to do the unmounts in PreRun and Teardown instead of DataIn and DataOut.

### DIFF
--- a/controllers/nnf_workflow_controller_helpers.go
+++ b/controllers/nnf_workflow_controller_helpers.go
@@ -691,9 +691,11 @@ func (r *NnfWorkflowReconciler) unmountNnfAccessIfNecessary(ctx context.Context,
 		},
 	}
 
+	// Check if the NnfAccess exists. If it doesn't, then we don't need to unmount anything. If we didn't
+	// find it in the cache and it really does exist, we'll eventually get an event from the API server and
+	// do the unmount then.
 	if err := r.Get(ctx, client.ObjectKeyFromObject(access), access); err != nil {
-		err = fmt.Errorf("Could not get NnfAccess %v: %w", client.ObjectKeyFromObject(access), err)
-		return nil, nnfv1alpha1.NewWorkflowError("Unable to find compute node mount information").WithError(err)
+		return nil, nil
 	}
 
 	teardownState, found := access.Labels[nnfv1alpha1.DataMovementTeardownStateLabel]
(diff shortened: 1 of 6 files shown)


controllers/nnf_workflow_controller_helpers.go
```diff
@@ -691,9 +691,11 @@ func (r *NnfWorkflowReconciler) unmountNnfAccessIfNecessary(ctx context.Context,
 		},
 	}
 
+	// Check if the NnfAccess exists. If it doesn't, then we don't need to unmount anything. If we didn't
+	// find it in the cache and it really does exist, we'll eventually get an event from the API server and
+	// do the unmount then.
 	if err := r.Get(ctx, client.ObjectKeyFromObject(access), access); err != nil {
-		err = fmt.Errorf("Could not get NnfAccess %v: %w", client.ObjectKeyFromObject(access), err)
-		return nil, nnfv1alpha1.NewWorkflowError("Unable to find compute node mount information").WithError(err)
+		return nil, nil
```
NateThornton, line 698: to be more precise, should this be `return nil, errors.IgnoreIfNotFound(err)`?

matthew-richerson: > to be more precise, should this be `return nil, errors.IgnoreIfNotFound(err)`?

Yes. Will change

controllers/nnf_workflow_controller_helpers.go
```diff
@@ -691,9 +691,11 @@ func (r *NnfWorkflowReconciler) unmountNnfAccessIfNecessary(ctx context.Context,
 		},
 	}
 
+	// Check if the NnfAccess exists. If it doesn't, then we don't need to unmount anything. If we didn't
```
NateThornton, line 694: It's kinda too bad we're going to be doing this for all workflows, not just data movements. I wonder if it's a performance concern.

matthew-richerson: > It's kinda too bad we're going to be doing this for all workflows, not just data movements. I wonder if it's a performance concern.

It's just doing a Get() on the client cache, so I don't think there's a performance issue.

ajfloeder (approved):

bdevcich-hpe (approved):

PR #139, opened 2023-02-01
RABSW-1128: Make fake mounts on kind Rabbit nodes

matthew-richerson:
Create empty directories on the Rabbit nodes in the clientmount reconciler to better fake out data movement and user containers.

### DIFF
--- a/main.go
+++ b/main.go
@@ -192,23 +192,12 @@ func (c *nodeLocalController) SetupReconcilers(mgr manager.Manager, opts *nnf.Op
 		return err
 	}
 
-	if os.Getenv("ENVIRONMENT") == "kind" {
-		if err := (&controllers.NnfClientMountReconciler{
-			Client:    mgr.GetClient(),
-			Log:       ctrl.Log.WithName("controllers").WithName("NnfClientMount"),
-			Scheme:    mgr.GetScheme(),
-			FakeMount: true,
-		}).SetupWithManager(mgr); err != nil {
-			return err
-		}
-	} else {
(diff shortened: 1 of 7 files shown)


main.go
```diff
@@ -192,7 +192,16 @@ func (c *nodeLocalController) SetupReconcilers(mgr manager.Manager, opts *nnf.Op
 		return err
 	}
 
-	if os.Getenv("ENVIRONMENT") != "kind" {
```
NateThornton, line 195: If I'm reading this right, you can just pass the if condition to the FakeMount param
```go
		if err := (&controllers.NnfClientMountReconciler{
			Client:    mgr.GetClient(),
			Log:       ctrl.Log.WithName("controllers").WithName("NnfClientMount"),
			Scheme:    mgr.GetScheme(),
			FakeMount: os.Getenv("ENVIRONMENT") == "kind",
		}).SetupWithManager(mgr); err != nil {
			return err
		}
```

NateThornton: Or, there appears to be only one spot where FakeMount is used, and it could be replaced by the conditional `if os.Getenv("ENVIRONMENT") == "kind" {` like you did in the nnf_access_controller. I prefer this instead of expanding the struct since searching for `== "kind"` would be consistent.

NateThornton (approved):

main.go
matthew-richerson: Sounds good. I'll change it

NateThornton (approved):

NateThornton (approved):

bdevcich-hpe (approved):

PR #142, opened 2023-02-06
Support for non-hardcoded container profiles

NateThornton:
- Support for non-hardcoded container profiles
- Remove pinned container profile creation out of the directive breakdown and to the workflow


config/samples/nnf_v1alpha1_nnfcontainerprofile.yaml
```diff
@@ -37,8 +27,19 @@ data:
...
+data:
+  template:
+    spec:
+      containers:
+      - name: forever
+        image: alpine:latest
+        command:
+        - /bin/sh
+        - -c
+        - "while true; do date && sleep 5; done"
```
bdevcich-hpe, line 45: Did you mean to check this in?

NateThornton: Yes. It's a good example of a container that never exits. Hence the name 'forever'

controllers/nnf_workflow_controller_helpers.go
```diff
@@ -1089,7 +1097,7 @@ func (r *NnfWorkflowReconciler) createOrUpdateContainerJobsIfNecessary(ctx conte
 				podSpec.RestartPolicy = corev1.RestartPolicyNever
 
 				// Because the IP changes, we need to have deterministic hostname for the pod
-				podSpec.Hostname = jobName
+				podSpec.Hostname = nnfNode
```
bdevcich-hpe, line 1100: This should be ok since the service name (subdomain) is the workflow name, right?

controllers/nnf_workflow_controller_helpers.go
```diff
@@ -1232,3 +1239,89 @@ func (r *NnfWorkflowReconciler) checkContainerJobs(ctx context.Context, workflow
...
+	}
+
+	if err := r.Get(ctx, client.ObjectKeyFromObject(profile), profile); err != nil {
+		return nil, err
+	}
+
+	return profile, nil
+}
+
+func (r *NnfWorkflowReconciler) createPinnedContainerProfileInNecessary(ctx context.Context, workflow *dwsv1alpha1.Workflow, index int) error {
```
roehrich-hpe, line 1283: Maybe this function name has a typo.  Should that "In" be "If" ?

config/samples/nnf_v1alpha1_nnfcontainerprofile.yaml
NateThornton: The container profile should not define the security context or tolerations since those are values populated by Rabbit software.

I will restore the storages.

bdevcich-hpe (approved):

controllers/nnf_workflow_controller_helpers.go
NateThornton: The pod and service are created in the workflow namespace, so that should be `nnf-node-name.service-name.workflow-namespace.svc.cluster.local`

controllers/directivebreakdown_controller.go
```diff
@@ -276,17 +276,6 @@ func (r *DirectiveBreakdownReconciler) Reconcile(ctx context.Context, req ctrl.R
 		if err != nil {
 			return ctrl.Result{}, err
 		}
-	case "container":
-		pinnedContainerProfile, err := createPinnedContainerProfile(ctx, r.Client, r.Scheme, argsMap, dbd, commonResourceName)
```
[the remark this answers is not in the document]
NateThornton: Because a container profile doesn't need a directive breakdown. I can move it to the proposal phase if you like. It just doesn't make sense to me that proposal would _do_ stuff. But it's just a meaningless name at this point.

matthew-richerson (approved):

controllers/directivebreakdown_controller.go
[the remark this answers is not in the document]
NateThornton: Moved to Proposal stage

PR #143, opened 2023-02-07
Mount container storages into containers via volumes

bdevcich-hpe:
Also use Create instead of CreateOrUpdate for Container Jobs/Service

### DIFF
--- a/controllers/nnf_workflow_controller_helpers.go
+++ b/controllers/nnf_workflow_controller_helpers.go
@@ -984,17 +984,7 @@ func (r *NnfWorkflowReconciler) removeAllPersistentStorageReferences(ctx context
 	return nil
 }
 
-// func (r *NnfWorkflowReconciler) createOrUpdateContainerServiceIfNecessary(ctx context.Context, workflow *dwsv1alpha1.Workflow) error {
-
-// 	log := log.FromContext(ctx)
-// 	log.Info("TODO: createOrUpdateContainerServiceIfNecessary() not implemented")
-
-// 	return nil
-// }
-
 func (r *NnfWorkflowReconciler) createContainerService(ctx context.Context, workflow *dwsv1alpha1.Workflow) error {
-	log := log.FromContext(ctx)
-
 	service := &corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      workflow.Name,
@@ -1017,17 +1007,12 @@ func (r *NnfWorkflowReconciler) createContainerService(ctx context.Context, work
 		if !apierrors.IsAlreadyExists(err) {
 			return err
 		}
-		log.Info("Service already exists", "object", client.ObjectKeyFromObject(service).String())
 	}
-
-	log.Info("Created Service", "object", client.ObjectKeyFromObject(service).String())
 
 	return nil
 }
 
 func (r *NnfWorkflowReconciler) createContainerJobs(ctx context.Context, workflow *dwsv1alpha1.Workflow, dwArgs map[string]string, index int) error {
-	log := log.FromContext(ctx)
-
 	profile, err := r.findPinnedContainerProfile(ctx, workflow, index)
 	if err != nil {
 		return err
@@ -1062,10 +1047,13 @@ func (r *NnfWorkflowReconciler) createContainerJobs(ctx context.Context, workflo
 		dwsv1alpha1.InheritParentLabels(job, workflow)
 		dwsv1alpha1.AddOwnerLabels(job, workflow)
 		dwsv1alpha1.AddWorkflowLabels(job, workflow)
-		addPinnedContainerProfileLabel(job, profile)
+
 		labels := job.GetLabels()
 		labels[nnfv1alpha1.ContainerLabel] = workflow.Name
+		labels[nnfv1alpha1.PinnedContainerProfileLabelName] = profile.GetName()
+		labels[nnfv1alpha1.PinnedContainerProfileLabelNameSpace] = profile.GetNamespace()
 		job.SetLabels(labels)
+
 		if err := ctrl.SetControllerReference(workflow, job, r.Scheme); err != nil {
 			return nnfv1alpha1.NewWorkflowErrorf("setting Job controller reference failed for '%s':", job.Name).WithError(err)
 		}
@@ -1095,9 +1083,7 @@ func (r *NnfWorkflowReconciler) createContainerJobs(ctx context.Context, workflo
 			if !apierrors.IsAlreadyExists(err) {
 				return err
 			}
-			log.Info("Job already exists", "object", client.ObjectKeyFromObject(job).String())
 		}
-		log.Info("Created Job", "object", client.ObjectKeyFromObject(job).String())
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


controllers/nnf_workflow_controller_helpers.go
```diff
@@ -1057,106 +1058,50 @@ func (r *NnfWorkflowReconciler) createOrUpdateContainerJobsIfNecessary(ctx conte
...
+			log.Info("Job already exists", "object", client.ObjectKeyFromObject(job).String())
 		}
+		log.Info("Created Job", "object", client.ObjectKeyFromObject(job).String())
 	}
 
 	return nil
 }
 
+// Retrieve the computes for the workflow and find their local nnf nodes
 func (r *NnfWorkflowReconciler) getNnfNodesForContainers(ctx context.Context, workflow *dwsv1alpha1.Workflow) ([]string, error) {
```
ajfloeder, line 1104: I surprised that containers is the only area that we need to know the list of Rabbits associated with the compute nodes. Just thinking this could named, getNnfNodesForComputes.

controllers/nnf_workflow_controller_helpers.go
```diff
@@ -1057,106 +1058,50 @@ func (r *NnfWorkflowReconciler) createOrUpdateContainerJobsIfNecessary(ctx conte
...
+		// Set the job's podspec from the container profile and add in volumes
+		r.setContainerJobPodSpec(ctx, workflow.Name, job, *profile, nnfNode, volumes)
 
-			return nil
+		// This is a timeout for the job, if set, when the deadline hits, the job will kill all
+		// pods and fail the job.
+		if profile.Data.ActiveDeadlineSeconds > 0 {
+			job.Spec.ActiveDeadlineSeconds = &profile.Data.ActiveDeadlineSeconds
 		}
+		// This defaults to 6 and is the maximum number of pod started before considering the
```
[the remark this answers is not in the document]
bdevcich-hpe: In our case, we're using a restart policy of `Never`. Using `Never` actually causes a new pod to start rather than trying to restart one that failed. The reasoning for this is to ensure the pod history/logs are easily found when pods fail.

So this number means that it will retry 6 times. In other words, retry (by starting a new pod) up to 6 times before the job stops firing up new pods and considers the job failed.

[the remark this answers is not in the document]
ajfloeder: Maybe you could give a link to documentation here that I could read to get the information you just put in this review comment? :)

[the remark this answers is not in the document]
bdevcich-hpe: Here's a comment I left in the container PR to explain this to Nate. It's confusing: https://github.com/NearNodeFlash/NearNodeFlash.github.io/pull/26#discussion_r1089460308. There some links to documentation in there as well.

I'd be happy to chat about this as well. It might be easier to explain over the phone.

controllers/nnf_workflow_controller_helpers.go
bdevcich-hpe: Good call. I renamed it.

controllers/nnf_workflow_controller_helpers.go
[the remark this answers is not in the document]
ajfloeder: Maybe put that link into this comment?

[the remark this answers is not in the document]
bdevcich-hpe: Will do.

controllers/nnf_workflow_controller_helpers.go
```diff
@@ -996,35 +1002,30 @@ func (r *NnfWorkflowReconciler) createOrUpdateContainerServiceIfNecessary(ctx co
...
 	}
 
-	result, err := ctrl.CreateOrUpdate(ctx, r.Client, service, mutateFn)
+	err := r.Create(ctx, service)
 	if err != nil {
-		return err
+		if !apierrors.IsAlreadyExists(err) {
+			return err
+		}
+		log.Info("Service already exists", "object", client.ObjectKeyFromObject(service).String())
```
NateThornton, line 1020: Is this log helpful? We should try to keep things lean. Maybe log.Debug()?

bdevcich-hpe: We only have Info() and Error(). I thought Info() was better than nothing, but I can take it out.

ajfloeder (approved):

controllers/nnf_workflow_controller_helpers.go
```diff
@@ -1325,3 +1273,155 @@ func (r *NnfWorkflowReconciler) createPinnedContainerProfileIfNecessary(ctx cont
 
 	return nil
 }
+
+func addPinnedContainerProfileLabel(object metav1.Object, nnfContainerProfile *nnfv1alpha1.NnfContainerProfile) {
```
NateThornton, line 1277: Being only used once and being brief, I'm not sure I'd this as a function.

controllers/nnf_workflow_controller_helpers.go
```diff
@@ -1325,3 +1273,155 @@ func (r *NnfWorkflowReconciler) createPinnedContainerProfileIfNecessary(ctx cont
...
+		labels = make(map[string]string)
+	}
+
+	labels[nnfv1alpha1.PinnedContainerProfileLabelName] = nnfContainerProfile.GetName()
+	labels[nnfv1alpha1.PinnedContainerProfileLabelNameSpace] = nnfContainerProfile.GetNamespace()
+	object.SetLabels(labels)
+}
+
+// Create a list of volumes to be mounted inside of the containers based on the DW_JOB/DW_PERSISTENT arguments
+func (r *NnfWorkflowReconciler) getContainerVolumes(ctx context.Context, workflow *dwsv1alpha1.Workflow, dwArgs map[string]string, nnfNodeName string) ([]nnfContainerVolume, error) {
```
NateThornton, line 1289: @matthew-richerson Aren't all the Rabbit mounts located at the same path? I thought we'd be able to get the NNF Access that describes the mount once, and use that mountpoint for all container volumes.

controllers/nnf_workflow_controller_helpers.go
```diff
@@ -1057,106 +1058,53 @@ func (r *NnfWorkflowReconciler) createOrUpdateContainerJobsIfNecessary(ctx conte
 			},
 		}
 
-		// Apply Job Labels
-		// TODO: confirm labels here; any missing?
+		// Apply Job Labels/Owners
```
NateThornton, line 1061: Optimizations:
* 99% of the job spec is the same across all jobs. You should create one structure up front, apply labels, volumes, etc. I think only the nodeSelector actually needs to be modified inside the loop.
* Watch your logs inside these loops: You may see only 1 log today, but that'll be hundreds in production.

controllers/nnf_workflow_controller_helpers.go
NateThornton: That's right, logr supports V() levels and not Debug. We don't have well-defined levels in our code, and I think now is not the time to introduce it.

But, as to the usefulness, it helps me to ask myself "Am I learning anything from the log statement?" Otherwise it's just noise.

controllers/nnf_workflow_controller_helpers.go
matthew-richerson: Yes, you could just look at the NnfAccess. LLNL wanted the index to compute node name translation for gfs2 file systems, and to do that we'll need to look at the individual ClientMount resources. So I think this code will be useful in the future.

bdevcich-hpe: I'll give this a go and only use the NnfAccess. Even though I talked to @matthew-richerson about this, I think I misunderstood what he was saying in the first place.

NateThornton: > LLNL wanted the index to compute node name translation for gfs2 file systems, and to do that we'll need to look at the individual ClientMount resources

I think they wanted links, which would be an enhancement to the NNF client mount controller to create when the mount occurs. Maybe I'm missing something, but I don't see container logic ever needing to get the client mounts.

controllers/nnf_workflow_controller_helpers.go
bdevcich-hpe: You're right. Only job `name`, `hostname`, and `nodeSelector` need to be changed. I've optimized the job creation loop to just share 1 job and then update those 3 fields for every NNF node (since they are based on the NNF Node name).

controllers/nnf_workflow_controller_helpers.go
bdevcich-hpe: I've updated this to use NNFAccess rather than ClientMount.

NateThornton (approved):

PR #145, opened 2023-02-07
RABSW-1139: Fix ClientMount directory create/remove for kind environment

matthew-richerson:
In the ClientMount controller for kind nodes, check whether the directory exists before creating or removing it.

Re-vendor dws

### DIFF
--- a/controllers/nnf_clientmount_controller.go
+++ b/controllers/nnf_clientmount_controller.go
@@ -197,13 +197,23 @@ func (r *NnfClientMountReconciler) changeMount(ctx context.Context, clientMountI
 
 	if os.Getenv("ENVIRONMENT") == "kind" {
 		if shouldMount {
+			// Return if the directory already exists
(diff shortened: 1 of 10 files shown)


NateThornton (approved):

ajfloeder (approved):

controllers/nnf_clientmount_controller.go
```diff
@@ -197,13 +197,23 @@ func (r *NnfClientMountReconciler) changeMount(ctx context.Context, clientMountI
 
 	if os.Getenv("ENVIRONMENT") == "kind" {
 		if shouldMount {
+			// Return if the directory already exists
+			if _, err := os.Stat(clientMountInfo.MountPath); !os.IsNotExist(err) {
```
bdevcich-hpe, line 201: Couldn't you use `os.IsExist()` here instead?

matthew-richerson: No, that checks for EEXIST which os.Stat() won't return. That made me look at this line again, though, and it's wrong for other reasons. Will fix

PR #148, opened 2023-02-09
Ensure 'key' values in the ruleset match against the exact string

NateThornton:

### DIFF
--- a/config/dws/nnf-ruleset.yaml
+++ b/config/dws/nnf-ruleset.yaml
@@ -8,23 +8,23 @@ spec:
   - command: "jobdw"
     watchStates: Proposal,Setup,PreRun,PostRun,Teardown
     ruleDefs:
-      - key: "type"
+      - key: "^type$"
         type: "string"
         pattern: "^(raw|xfs|gfs2|lustre)$"
         isRequired: true
         isValueRequired: true
-      - key: "capacity"
+      - key: "^capacity$"
         type: "string"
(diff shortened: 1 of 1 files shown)


ajfloeder (approved):

roehrich-hpe:
I misunderstood this when you mentioned it earlier.   Why would the key need to be a regex?

bdevcich-hpe (approved):

bdevcich-hpe:
> I misunderstood this when you mentioned it earlier. Why would the key need to be a regex?

To support storage arguments for the container directive. Example:

```
#DW container name=blake-container profile=nnfcontainerprofile-sample \
       DW_JOB_foo-local-storage=blake-gfs2 \
       DW_PERSISTENT_foo-persistent-storage=blake-persistent
```

We need to be able to pass in whatever we want after the `DW_JOB_` or `DW_PERSISTENT_`, so using a regex for the key allows us to be able to do that.

roehrich-hpe:
> > I misunderstood this when you mentioned it earlier. Why would the key need to be a regex?
> 
> To support storage arguments for the container directive. Example:
> 
> ```
> #DW container name=blake-container profile=nnfcontainerprofile-sample \
>        DW_JOB_foo-local-storage=blake-gfs2 \
>        DW_PERSISTENT_foo-persistent-storage=blake-persistent
> ```
> 
> We need to be able to pass in whatever we want after the `DW_JOB_` or `DW_PERSISTENT_`, so using a regex for the key allows us to be able to do that.

got it.  Thanks.

roehrich-hpe (approved):

PR #153, opened 2023-02-14
PostRun Ready Logic

bdevcich-hpe:
PostRun will go `Ready:true` only if all jobs/pods have finished successfully. If containers are still running, PostRun requeues every 2 seconds until the jobs/containers have indicated a finish (pass or fail).

Once containers have finished, the results are checked. If all jobs are successful, then PostRun goes `Ready:true`. If not, then it stays `Ready:false`. Jobs are successful when a child pod has completed successfully (i.e. exited non-zero).

Additionally, a check now occurs when creating `NnfAccesses` for the rabbits to see if the `NnfAccess` is referenced by a container directive. If so, the `teardownState` is set to Teardown to ensure that mounts are present during the potential life of the containers through the PostRun state. This case only occurs during the PostRun state when a pod fails and a retry is attempted.

Additionally, `backoffLimit` and `activeDeadlineSeconds` in the NnfContainerProfile have been renamed to `retryLimit` and `postRunTimeoutSeconds` to make them more meaningful to the user.

`postRunTimeoutSeconds` is used to set a timeout on the jobs/pods once PostRun is started. When the timeout is hit, the job will stop the pods and the job will finish with a timeout failure condition.

### DIFF
--- a/controllers/nnf_workflow_controller.go
+++ b/controllers/nnf_workflow_controller.go
@@ -845,11 +845,11 @@ func (r *NnfWorkflowReconciler) startPreRunState(ctx context.Context, workflow *
 			},
 		}
 
-		// Set the teardown state to post run. If there is a copy_out directive that uses
-		// this storage instance, set the teardown state so NNF Access is preserved through
+		// Set the teardown state to post run. If there is a copy_out or container directive that
+		// uses this storage instance, set the teardown state so NNF Access is preserved through
 		// DataOut.
 		teardownState := dwsv1alpha1.StatePostRun
-		if findCopyOutDirectiveIndexByName(workflow, dwArgs["name"]) >= 0 {
+		if findCopyOutDirectiveIndexByName(workflow, dwArgs["name"]) >= 0 || findContainerDirectiveIndexByName(workflow, dwArgs["name"]) >= 0 {
 			teardownState = dwsv1alpha1.StateTeardown
 		}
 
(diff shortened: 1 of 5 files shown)


controllers/nnf_workflow_controller_helpers.go
```diff
@@ -1243,36 +1243,107 @@ func (r *NnfWorkflowReconciler) waitForContainersToStart(ctx context.Context, wo
...
 		}
 	}
 
-	// TODO: Once postrun starts, we will want a timeout that starts and eventually kills/fails job pods
-	// once that timeout is hit.
+	return nil, nil
+}
+
+func (r *NnfWorkflowReconciler) getContainerJobs(ctx context.Context, workflow *dwsv1alpha1.Workflow, index int) (*batchv1.JobList, error) {
+	jobList := &batchv1.JobList{}
```
NateThornton, line 1320: Precedence has the lables (the configurable stuff) above the returned object `jobList`, above the r.List() method

So
```go
matchLabels := dwsv1alpha1.MatchingWorkflow(workflow)
matchLabels[nnfv1alpha1.DirectiveIndexLabel] = strconv.Itoa(index)

jobs := &batchv1.JobList{}
if err := r.List(ctx, jobList, matchLabels); err != nil {
   ...
}
```

NateThornton: Everywhere else we've wrapped matchLabels in a []client.ListOption array, but yours makes more sense. Might be worth a cleanup ticket sometime.

controllers/nnf_workflow_controller.go
```diff
@@ -845,11 +845,11 @@ func (r *NnfWorkflowReconciler) startPreRunState(ctx context.Context, workflow *
 			},
 		}
 
-		// Set the teardown state to post run. If there is a copy_out directive that uses
-		// this storage instance, set the teardown state so NNF Access is preserved through
+		// Set the teardown state to post run. If there is a copy_out or container directive that
+		// uses this storage instance, set the teardown state so NNF Access is preserved through
 		// DataOut.
```
ajfloeder, line 850: s/DataOut/Teardown

controllers/nnf_workflow_controller_helpers.go
```diff
@@ -1042,27 +1055,18 @@ func (r *NnfWorkflowReconciler) createContainerJobs(ctx context.Context, workflo
 	labels[nnfv1alpha1.PinnedContainerProfileLabelName] = profile.GetName()
 	labels[nnfv1alpha1.PinnedContainerProfileLabelNameSpace] = profile.GetNamespace()
 	labels[nnfv1alpha1.DirectiveIndexLabel] = strconv.Itoa(index)
+	labels[nnfv1alpha1.DirectiveIndexLabel] = strconv.Itoa(index)
```
ajfloeder, line 1058: Are you setting this twice on purpose?

controllers/nnf_workflow_controller_helpers.go
```diff
@@ -745,6 +746,26 @@ func findCopyOutDirectiveIndexByName(workflow *dwsv1alpha1.Workflow, name string
 	return -1
 }
 
+// Returns the directive index matching the container directive whose storage field(s) reference
+// the provided name argument, or -1 if not found.
+func findContainerDirectiveIndexByName(workflow *dwsv1alpha1.Workflow, name string) int {
+	for idx, directive := range workflow.Spec.DWDirectives {
+		if strings.HasPrefix(directive, "#DW container") {
```
matthew-richerson, line 753: You can use `parameters["command"] == container` instead of matching on "#DW container". That will prevent any whitespace differences from tripping things up.

matthew-richerson (approved):

controllers/nnf_workflow_controller_helpers.go
bdevcich-hpe: No, I think this was a merge artifact from when I rebased. Good catch.

NateThornton (approved):

PR #156, opened 2023-02-15
Create new examples folder that gets deployed after the webhook is ready

NateThornton:
I didn't like that we were hacking our way to deploy just the placeholder storage profile when we'll need the same code for containers. So I moved everything to a new examples folder and deploy that using kustomize
.


config/examples/nnf_v1alpha1_nnfcontainerprofiles.yaml
```diff
@@ -0,0 +1,41 @@
...
+data:
+  template:
+    spec:
+      containers:
+      - name: forever
+        image: alpine:latest
+        command:
+        - /bin/sh
+        - -c
+        - "while true; do date && sleep 5; done"
```
bdevcich-hpe, line 41: For testing, might be nice to have one more that exits cleanly after a while.

bdevcich-hpe (approved):

config/examples/nnf_v1alpha1_nnfcontainerprofiles.yaml
NateThornton: I agree, but that's beyond the scope of this PR.

For testing, there's currently a 60s timeout that the workflow advances. Keep that in mind if we exceed that as a result of a container profile

PR #157, opened 2023-02-16
github-151: Fix LVM issues with gfs2

matthew-richerson:
This commit fixes two issues that were affecting gfs2 file systems:
 - The dlm lock manager was failing to lock because the VG name was too long
 - The lvcreate command needs an "--activate ys" to active a shared volume

### DIFF
--- a/config/samples/placeholder_nnfstorageprofile.yaml
+++ b/config/samples/placeholder_nnfstorageprofile.yaml
@@ -28,7 +28,7 @@ data:
         deactivate: --activate n $VG_NAME
         lockStart: --lock-start $VG_NAME
       vgRemove: $VG_NAME
-      lvCreate: --activate ys -l 100%VG --stripes $DEVICE_NUM --stripesize=32KiB --name $LV_NAME $VG_NAME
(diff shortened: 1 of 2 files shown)


ajfloeder (approved):

config/samples/placeholder_nnfstorageprofile.yaml
```diff
@@ -28,7 +28,7 @@ data:
         deactivate: --activate n $VG_NAME
         lockStart: --lock-start $VG_NAME
       vgRemove: $VG_NAME
-      lvCreate: -l 100%VG --stripes $DEVICE_NUM --stripesize=32KiB --name $LV_NAME $VG_NAME
+      lvCreate: --activate ys -l 100%VG --stripes $DEVICE_NUM --stripesize=32KiB --name $LV_NAME $VG_NAME
```
NateThornton, line 31: This is unrelated, but while you're at it, can you change the `-l` to the longform `--extents` so we match other options?

This line and all lvcreate lines

NateThornton (approved):

bdevcich-hpe (approved):

PR #158, opened 2023-02-16
Prepare container profiles for int-test

bdevcich-hpe:
A few changes to aid in running int-test at containers. Updated the sample profile with some documentation. Removed the container empty block from teardown.

This will be the last PR against the feature branch.

I used these changes to write container tests for int-test. Those changes will be coming after I merge the feature branch.

### DIFF
--- a/config/examples/nnf_v1alpha1_nnfcontainerprofiles.yaml
+++ b/config/examples/nnf_v1alpha1_nnfcontainerprofiles.yaml
@@ -1,15 +1,33 @@
 apiVersion: nnf.cray.hpe.com/v1alpha1
 kind: NnfContainerProfile
 metadata:
-  name: randomly-fail
+  name: example-success
 data:
-  retryLimit: 6
-  # postRunTimeoutSeconds: 0
   storages:
     - name: DW_JOB_foo-local-storage
       optional: false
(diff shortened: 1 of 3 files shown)


config/examples/nnf_v1alpha1_nnfcontainerprofiles.yaml
```diff
@@ -1,15 +1,33 @@
...
-  # postRunTimeoutSeconds: 0
   storages:
     - name: DW_JOB_foo-local-storage
       optional: false
     - name: DW_PERSISTENT_foo-persistent-storage
+      optional: true
+  template:
+    spec:
+      containers:
+      - name: forever
```
NateThornton, line 14: I've been matching the container name with the profile name

NateThornton (approved):

config/examples/nnf_v1alpha1_nnfcontainerprofiles.yaml
bdevcich-hpe: Good call. I overlooked that.

NateThornton (approved):

matthew-richerson (approved):

PR #159, opened 2023-02-17
Land feature/containers

bdevcich-hpe:
All the code in here has been reviewed previously via individual Pull Requests.

Any thoughts on whether I should squash this into one commit or not?


config/dws/nnf-ruleset.yaml
```diff
@@ -92,3 +92,18 @@ spec:
         type: "string"
         isRequired: true
         isValueRequired: true
+  - command: "container"
+    watchStates: Proposal,PreRun,PostRun,Teardown
+    ruleDefs:
+      - key: "name"
```
NateThornton, line 98: This and the "profile" key below require the regexp `'^name$'` and `'^profile$'`, respectively

config/samples/nnf_v1alpha1_nnfcontainerprofile.yaml
```diff
@@ -0,0 +1,44 @@
+apiVersion: nnf.cray.hpe.com/v1alpha1
```
NateThornton, line 1: Delete or move to ./config/examples if you plan to use it

controllers/nnf_workflow_controller.go
```diff
@@ -82,12 +82,12 @@ type NnfWorkflowReconciler struct {
 //+kubebuilder:rbac:groups=dws.cray.hpe.com,resources=computes,verbs=get;create;list;watch;update;patch
 //+kubebuilder:rbac:groups=cray.hpe.com,resources=lustrefilesystems,verbs=get;list;watch
 
+//+kubebuilder:rbac:groups=nnf.cray.hpe.com,resources=nnfcontainerprofiles,verbs=get;create;list;watch;update;patch;delete;deletecollection
+//+kubebuilder:rbac:groups=batch,resources=jobs,verbs=get;list;watch;create;update;patch;delete;deletecollection
```
NateThornton, line 86: You don't have a watch on jobs so the controller won't receive Reconcile()s. Is this OK because you are polling?

config/samples/nnf_v1alpha1_nnfcontainerprofile.yaml
bdevcich-hpe: My thought was to keep this as a fully documented sample with all of the options that could be used as a reference. I don't plan on this one getting deployed like the ones in the examples folder.

NateThornton: In that case can you make the spec.containers.command less insane

config/samples/nnf_v1alpha1_nnfcontainerprofile.yaml
```diff
@@ -0,0 +1,44 @@
...
+  namespace: nnf-system
+data:
+  # Specifies the number of times a container will be retried upon a failure. A
+  # new pod is deployed on each retry. Defaults to 6 by kubernetes itself and
+  # must be set. A value of 0 disables retries.
+  retryLimit: 6
+
+  # Stop any containers after X seconds once a workflow has transitioned to
+  # PostRun. A value of 0 disables this behavior.
+  postRunTimeoutSeconds: 0
```
NateThornton, line 14: Does this have a default?

controllers/nnf_workflow_controller.go
NateThornton: Then do you need the watch? I guess other resources aren't all that tidy either. So maybe leave it.

controllers/nnf_workflow_controller.go
```diff
@@ -213,13 +213,14 @@ func (r *NnfWorkflowReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		}
 
 		driverStatus.Status = dwsv1alpha1.StatusRunning
-		driverStatus.Message = ""
```
NateThornton, line 216: I'd leave this since it is clear you are clearing the block of driver status values.  Then, for non-nil result, overwrite the message field. Same number of lines of code, but the organization makes it a little more obvious (for me)

config/samples/nnf_v1alpha1_nnfcontainerprofile.yaml
bdevcich-hpe: It essentially defaults to 0 with omitempty since it's an int.

NateThornton (approved):

config/samples/nnf_v1alpha1_nnfcontainerprofile.yaml
NateThornton: If you are using this sample as documentation, please include the default values

controllers/nnf_workflow_controller.go
bdevcich-hpe: Doesn't seem like we do. I'll remove it.

config/samples/nnf_v1alpha1_nnfcontainerprofile.yaml
bdevcich-hpe: I'll make it just return 0 after 15 seconds.

controllers/nnf_workflow_controller.go
bdevcich-hpe: I'll put it back.

config/samples/nnf_v1alpha1_nnfcontainerprofile.yaml
bdevcich-hpe: Done

NateThornton (approved):

controllers/nnf_workflow_controller.go
```diff
@@ -840,18 +859,18 @@ func (r *NnfWorkflowReconciler) startPreRunState(ctx context.Context, workflow *
 		}
 	}
 
+	readyResult, err := r.waitForNnfAccessStateAndReady(ctx, workflow, index, "mounted")
```
matthew-richerson, line 862: Why did this get moved from finishPreRunState?

controllers/nnf_workflow_controller_helpers.go
```diff
@@ -849,3 +1002,484 @@ func (r *NnfWorkflowReconciler) removeAllPersistentStorageReferences(ctx context
 
 	return nil
 }
+
+func (r *NnfWorkflowReconciler) createContainerService(ctx context.Context, workflow *dwsv1alpha1.Workflow) error {
+	service := &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      workflow.Name,
```
matthew-richerson, line 1009: If there is more than one "#dw container" line, is it okay that you're only creating one service?

controllers/nnf_workflow_controller.go
bdevcich-hpe: I originally moved to this to startPreRunState so that when in finishPreRunState, we'd ensure that the NNFAccesses are mounted before we start checking to see if the containers started. But looking at this again, I don't think that is necessary. I'll put it back and test.

controllers/nnf_workflow_controller_helpers.go
bdevcich-hpe: Great question. I haven't test multiple container directives, which is something I will do in the future. Just having 1 service would allow all containers to talk to each other if need be. If we need to further partition this by the directive index and create multiple services, I think we could do that.

The service name ends up being the subdomain for the container (i.e. hostname.subdomain.namespace.svc.clusterlocal), so that could get a bit confusing.

controllers/nnf_workflow_controller.go
bdevcich-hpe: I remember now: keeping this in finishPreRunState at the top of the block ends up in an error because container directives do not have any NnfAccess. So I moved it to the end of startPreRunState because only non-container directives get that far - container directives return before doing anything with NnfAccess.

So I could keep this here or perhaps move it further down after `return r.waitForContainersToStart(ctx, workflow, index)`.

bdevcich-hpe: @matthew-richerson and I discussed this, and I kept this in `finishPreRunState()` and moved it down after the switch block where `r.waitForContainersToStart()` would return for container directives.

controllers/nnf_workflow_controller_helpers.go
bdevcich-hpe: I put in an a check in proposal for multiple container directives. For now, it will error out. The way things are currently working, if you supply multiple container directives you only get 1 job as well since it's based on the workflow name.

matthew-richerson (approved):

NateThornton (approved):

PR #163, opened 2023-02-24
NNF Port Manager

NateThornton:
Introduces a new resource, `NnfPortManager`, and controller of the same name. The NN Port Manager is designed to allocate ports, as provided in the system configuration, to any "requester" resource that might need them. The immediate need is to support customer containers.

I could type up a note on the usage of the port manager, but it's really pretty simple. If you want ports, add yourself to the `spec.Allocations`, then check for status in `status.Allocations`. When you're done, remove yourself.

Look to the tests as an example.

I did make some assumptions

- A valid System Configuration exists: The "Ports" value in the system configuration must be valid. That validation is deferred to another task (maybe a system configuration webhook?)
- Bad Port Manager clients don't exist: As we are the only client of the port manager, I took certain liberties with the client behavior: mainly that the client won't do evil shit like duplicate an existing allocation, or leave behind an allocation for eternity.
- Performance is ignored: The allocation and deallocation are of O(n^2) complexity. I don't really care to optimize this until we know the behavior.
- Reuse behavior. Ben went on about reusing ports "too soon" and needing a "TCP Port Reuse Delay. I didn't bother with that in detail, but we _do_ prefer unused ports over free'd ports. It's possible to expand on this and add delay timers and such, but I didn't bother for the first iteration.


api/v1alpha1/nnf_port_manager_types.go
```diff
@@ -0,0 +1,136 @@
+/*
+ * Copyright 2022 Hewlett Packard Enterprise Development LP
```
bdevcich-hpe, line 2: 2023

controllers/nnf_port_manager_controller_test.go
```diff
@@ -0,0 +1,184 @@
...
+		JustBeforeEach(func() {
+			Expect(k8sClient.Create(ctx, mgr)).To(Succeed())
+			DeferCleanup(func() { Expect(k8sClient.Delete(ctx, mgr)).To(Succeed()) })
+		})
+
+		reservePorts := func(mgr *nnfv1alpha1.NnfPortManager, name string, count int) {
+			By(fmt.Sprintf("Reserving %d ports for '%s'", count, name))
+
+			allocation := nnfv1alpha1.NnfPortManagerAllocationSpec{
+				Requester: corev1.ObjectReference{Name: name},
```
bdevcich-hpe, line 92: Normally, the reference object would have to actually exist, right?

api/v1alpha1/nnf_port_manager_types.go
```diff
@@ -0,0 +1,136 @@
...
+type NnfPortManagerAllocationStatus struct {
+	// Requester is an object reference to the requester of the port resource, if one exists, or
+	// empty otherwise.
+	Requester *corev1.ObjectReference `json:"requester,omitempty"`
+
+	// Ports is list of ports allocated to the owning resource.
+	Ports []uint16 `json:"ports,omitempty"`
+
+	// Status is the ownership status of the port.
+	Status NnfPortManagerAllocationStatusStatus `json:"status"`
```
bdevcich-hpe, line 84: Would it be better to call this a state instead of having the double status?

controllers/nnf_port_manager_controller.go
```diff
@@ -0,0 +1,296 @@
...
+	// in the status' until we have the desired number of ports.
+
+	const (
+		exhausted = iota
+		more
+	)
+
+	// Search for a free allocation and claim ports from that allocation. Returns 'more' if there
+	// are potentially more allocations available with free ports, and 'exhausted' otherwise.
+	claimPortsFromFreeAllocation := func() int {
```
bdevcich-hpe, line 257: Couldn't you just return a bool here?

bdevcich-hpe (approved):

controllers/nnf_port_manager_controller_test.go
NateThornton: I'm not sure I understand your question. As long as you can make an ObjectReference{}, it will work. Create one out of thin air if you want.

controllers/nnf_port_manager_controller.go
NateThornton: Then you get into the argument of what your return value represents. 

Does `claimPortsFromFreeAllocation() == true` mean I've claimed the ports I need? Does it mean I've exhausted the ports and failed?

With an enum like this, I've effectively implemented named bools.

api/v1alpha1/nnf_port_manager_types.go
NateThornton: We use 'Status' everywhere, so I prefer it this way. If this were any other language I'd used a nested class.

controllers/nnf_port_manager_controller_test.go
bdevcich-hpe: Ah. Initially I was thinking it has to reference an object that exists.

PR #164, opened 2023-02-28
Add MPI Hostfile support to NnfDataMovementCommandStatus

bdevcich-hpe:


NateThornton:
Why is this needed?

bdevcich-hpe:
For testing, for the user to understand what is contained inside the hostfile without having to shell into a pod and for being able to delete the hostfile when the NnfDataMovement is deleted.

NateThornton:
Why would the host file be anything besides the list of Rabbits and slot count? What value is this adding? These are resources stored in a database; while a host file isn't megabytes in size, any amount makes a difference.

Why would the host file need to linger any longer than the mpirun command? Because it is deterministic, it isn't required for debugging (it can be easily reconstituted from existing resources).

bdevcich-hpe:
> Why would the host file be anything besides the list of Rabbits and slot count? What value is this adding? These are resources stored in a database; while a host file isn't megabytes in size, any amount makes a difference.
> 
> Why would the host file need to linger any longer than the mpirun command? Because it is deterministic, it isn't required for debugging (it can be easily reconstituted from existing resources).

Because sometimes debugging is worth it's weight in gold. Since this can impact performance, I'll remove it.

PR #166, opened 2023-03-01
Nnf ec enhanced logging

NateThornton:
- Advance nnf-ec to the latest master which has added logging capabilities
- Use the controller logger, with name `ec`, as the nnf-ec logger.
- Convert zap usage to allow for command line parameters, while maintaining the current defaults.


controllers/nnf_node_ec_data_controller.go
```diff
@@ -83,8 +86,19 @@ func (r *NnfNodeECDataReconciler) Start(ctx context.Context) error {
 		persistent.StorageProvider = r
 	}
 
+	// NNF Element Controller logger provides 4 verbosity levels. By default, nnf-ec
+	// will use the raw logger provided. If you are using zap defaults, this will
+	// only include the INFO and DEBUG levels. You can increase the defaults, by
+	// manually setting the zap.Options.Level, or using the -zap-log-level flag.
+	//
+	// You can decrease the log level of nnf-ec by adding verbosity levels to passed
```
ajfloeder, line 94: Should this be, "by adding a verbosity level that is passed into the logger"?

main.go
```diff
@@ -86,15 +85,16 @@ func main() {
 
 	nnfopts := nnf.BindFlags(flag.CommandLine)
 
-	opts := zapcr.Options{
+	opts := &zap.Options{
 		Development: true,
+		TimeEncoder: zapcore.ISO8601TimeEncoder,
```
ajfloeder, line 90: Do you expect this to make all logs report in the same format?

NateThornton: No. It's just a means to default the zap time encoder to ISO8601 instead of hardcoding it.

controllers/nnf_node_ec_data_controller.go
NateThornton: I rewrote that line - please take a look and see if it makes sense this time around

ajfloeder (approved):

matthew-richerson (approved):

PR #167, opened 2023-03-06
Containers: Check for XFS/Raw filesystems

bdevcich-hpe:
These filesystems can only be mounted once - they are not supported for
containers.

### DIFF
--- a/controllers/nnf_workflow_controller_helpers.go
+++ b/controllers/nnf_workflow_controller_helpers.go
@@ -250,18 +250,34 @@ func (r *NnfWorkflowReconciler) validateContainerDirective(ctx context.Context, 
 		return fmt.Errorf("storage '%s' not found in container profile '%s'", storageName, profile.Name)
 	}
 
+	checkContainerFs := func(idx int) error {
+		getDirectiveFsType := func(idx int) string {
+			args, _ := dwdparse.BuildArgsMap(workflow.Spec.DWDirectives[idx])
+			return args["type"]
+		}
+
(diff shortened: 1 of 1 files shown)


ajfloeder (approved):

controllers/nnf_workflow_controller_helpers.go
```diff
@@ -250,19 +250,35 @@ func (r *NnfWorkflowReconciler) validateContainerDirective(ctx context.Context,
 		return fmt.Errorf("storage '%s' not found in container profile '%s'", storageName, profile.Name)
 	}
 
+	checkContainerFs := func(idx int) error {
+		getDirectiveFsType := func(idx int) string {
+			args, _ := dwdparse.BuildArgsMap(workflow.Spec.DWDirectives[idx])
+			return args["type"]
+		}
+
+		if t := getDirectiveFsType(idx); strings.ToLower(t) == "raw" || strings.ToLower(t) == "xfs" {
```
matthew-richerson, line 259: Maybe check the inverse of this instead? `if type == "lustre" || type == "gfs2"`. That makes it less likely to cause a bug if we add a new file system type in the future. It's up to you, though. We probably won't add anymore FS types.

matthew-richerson (approved):

controllers/nnf_workflow_controller_helpers.go
bdevcich-hpe: I actually had it this way to start and then waffled and switched to this. I'll change it back to look for lustre/gfs2.

ajfloeder (approved):

PR #172, opened 2023-03-17
Added NnfContainerProfile validation webhook + unit tests

bdevcich-hpe:
- Added validation webhook for container profiles
- Fix a bug in the container filesystem check for persistent filesystems
- Add unit tests for container directives, most notably the storages in the profile and in the container directive arguements
- Add integration test to ensure that targeted compute nodes select the correct local NNF nodes for container workflows


api/v1alpha1/nnfcontainerprofile_webhook.go
```diff
@@ -0,0 +1,90 @@
...
+			nnfcontainerprofilelog.Error(err, "invalid")
+			return err
+		}
+	}
+
+	return r.validateContent()
+}
+
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
+func (r *NnfContainerProfile) ValidateDelete() error {
```
matthew-richerson, line 73: Is this actually used? The Kubebuilder line above only lists create and update

api/v1alpha1/nnfcontainerprofile_webhook.go
```diff
@@ -0,0 +1,90 @@
...
+	// TODO(user): fill in your validation logic upon object deletion.
+	return nil
+}
+
+func (r *NnfContainerProfile) validateContent() error {
+	if r.Data.RetryLimit < 0 {
+		return fmt.Errorf("retryLimit cannot be less than 0")
+	}
+
+	if r.Data.PostRunTimeoutSeconds < 0 {
```
matthew-richerson, line 85: Does this do something that `kubebuilder:validation:Minimum` doesn't?

matthew-richerson (approved):

controllers/nnf_workflow_controller_test.go
```diff
@@ -101,6 +120,70 @@ var _ = Describe("NNF Workflow Unit Tests", func() {
...
+					Error:           "",
+					AllocationCount: 0,
+				}},
+			},
+		}
+		Expect(k8sClient.Create(context.TODO(), nnfStorage)).To(Succeed())
+	}
+
+	deletePersistentStorageInstance := func(name string) {
+		By("Fabricate the nnfStorage as it the persistent storage instance exists")
```
ajfloeder, line 170: s/as it/as if/

ajfloeder (approved):

api/v1alpha1/nnfcontainerprofile_webhook.go
bdevcich-hpe: No - I've removed this and added `kubebuilder:validation:Minimum` to these two values. Webhook gets even simplier then, but I assume we'll have more things to add in the future.

PR #176, opened 2023-03-22
Github #39: Separate NnfAccess mount/unmount code paths

matthew-richerson:
This commit separates out the logic for mounting and unmounting an NnfAccess. This was to provide proper unlocking of the NnfStorage for XFS and raw allocations.

### DIFF
--- a/controllers/nnf_access_controller.go
+++ b/controllers/nnf_access_controller.go
@@ -170,24 +170,19 @@ func (r *NnfAccessReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{Requeue: true}, nil
 	}
 
+	var result *ctrl.Result = nil
+
 	if access.Status.State == "mounted" {
-		result, err := r.mount(ctx, access, clientList, storageMapping)
-		if err != nil {
-			return ctrl.Result{}, err
-		}
-
-		if result != nil {
-			return *result, nil
-		}
+		result, err = r.mount(ctx, access, clientList, storageMapping)
 	} else {
-		result, err := r.unmount(ctx, access, clientList, storageMapping)
-		if err != nil {
-			return ctrl.Result{}, err
-		}
+		result, err = r.unmount(ctx, access, clientList, storageMapping)
+	}
 
-		if result != nil {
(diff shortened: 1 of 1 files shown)


controllers/nnf_access_controller.go
```diff
@@ -170,65 +170,113 @@ func (r *NnfAccessReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 	}
 
 	// Aggregate the status from all the ClientMount resources
 	ready, err = r.getClientMountStatus(ctx, access, clientList)
 	if err != nil {
-		return ctrl.Result{}, err
+		return nil, err
 	}
 
 	// Wait for all of the ClientMounts to be ready before setting the Ready field
```
ajfloeder, line 243: You're not directly setting the Ready field here, maybe a comment update?

controllers/nnf_access_controller.go
```diff
@@ -170,65 +170,113 @@ func (r *NnfAccessReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
-		}
+	// Aggregate the status from all the ClientMount resources
+	ready, err := r.getClientMountStatus(ctx, access, clientList)
+	if err != nil {
+		return nil, err
 	}
 
-	access.Status.Ready = true
-	access.Status.Error = nil
+	// Wait for all of the ClientMounts to be ready before setting the Ready field
```
ajfloeder, line 264: No longer setting Ready in this function, maybe an updated comment here too?

controllers/nnf_access_controller.go
```diff
@@ -170,65 +170,113 @@ func (r *NnfAccessReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		}
+	} else {
+		result, err := r.unmount(ctx, access, clientList, storageMapping)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+
+		if result != nil {
+			return *result, nil
+		}
```
bdevcich-hpe, line 190: Could you move these if tests outside the block and handle it once for both cases?

bdevcich-hpe (approved):

PR #177, opened 2023-04-05
Add MPI support to containers via mpi-operator

bdevcich-hpe:
This adds in a new way to create containers using mpi-operator. mpi-operator is now a requirement of nnf-sos in order to run MPI containers.

Users can now launch MPI container workflows. This is done via the NnfContainerProfile. Container workflows can be executed in two ways:
  - MPI (launcher/worker model)
  - Non-MPI (one command for all containers)

The launcher/worker model allows users to run `mpirun` on the launcher pod and then use the workers as nodes `mpirun`. See the mpi-operator docs for more: https://www.kubeflow.org/docs/components/training/mpi/.

Major Changes:

- Added `MPISpec` to define `MPIJobs` to container profile
- Moved original container implementation from `Template` to `Spec` to mimic `MPISpec` name. User now only defines the PodSpec rather than the PodTemplateSpec
- Added example-mpi NnfContainerProfile (used for testing)
- Added permissions to both MPI and non-MPI containers to run as non root users (i.e. `user` or `mpiuser`).
- Reworked PreRun to create either type and watch for successful container start for Ready logic.
- Reworked PostRun to watch for completion for either type and determine Ready state if containers completed successfuly.
- Added InitContainers to map the `user` or `mpiuser` to the workflow's User and Group ID. This allows ssh to work properly for mpirun.
- New functions added to support both MPI and non-MPI container creation logic.
- Use server-side deployment to workaround MPIJob's large CRD annotations


matthew-richerson (approved):

ajfloeder (approved):

roehrich-hpe (approved):

bdevcich-hpe:
Forced pushed to rebase on top of master

PR #180, opened 2023-04-18
Add support for extra dcp and dryrun options in NnfDataMovementSpec

bdevcich-hpe:
In order to support per-DM configuration options, we need to add some options to the spec. These values will override/supplement the existing data movement configuration that is defined in the nnf-dm-config ConfigMap.

In this case, LLNL has a need to add extra dcp options for a given data movement request. This will be done via the Copy Offload API, but we need to way to save this to the NnfDataMovement so it can be used when data movement is performed.

For debugging purposes, the dryrun option has also been added to fake out data movement.


api/v1alpha1/nnf_datamovement_types.go
```diff
@@ -72,6 +77,18 @@ type NnfDataMovementSpecSourceDestination struct {
 	StorageReference corev1.ObjectReference `json:"storageReference,omitempty"`
 }
 
+// NnfDataMovementConfig provides a way for a user to override the data movement behavior on a
+// per DM basis.
+type NnfDataMovementConfig struct {
+
+	// Fake the Data Movement operation. The system "performs" Data Movement but the command to so
```
ajfloeder, line 84:
```suggestion
	// Fake the Data Movement operation. The system "performs" Data Movement but the command to do so
```

ajfloeder (approved):

matthew-richerson (approved):

roehrich-hpe (approved):

PR #182, opened 2023-04-25
RABSW-1159: Update deploy.sh to look at the deployment ready count

matthew-richerson:
The deploy.sh was looking for a "1/1" ready field for the dws webhook. There may not be enough worker nodes on the system to run all 3 DWS webhooks, so some of the webhook pods may not be ready. If one of these pods shows up first in the pod list, then the deploy.sh script will hang forever. Instead, look at the number of ready replicas in the dws webhook deployment to be one or more.


roehrich-hpe (approved):
Thanks for fixing my mistake.

PR #184, opened 2023-04-25
RABSW-1158: Update nnf-ec and add timeout environment variable

matthew-richerson:
Make use of the new timeout in nnf-ec when running commands. Timeout commands after 90 seconds and return an error.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -44,8 +44,8 @@ spec:
           # use this node to create storage during setup phase
           - name: RABBIT_NODE
             value: "0"
-          - name: NNF_EC_COMMAND_TIMEOUT
-            value: "90s"
+          - name: NNF_EC_COMMAND_TIMEOUT_SECONDS
+            value: "90"
           - name: NNF_NODE_NAME
             valueFrom:
               fieldRef:
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -44,6 +44,8 @@ spec:
           # use this node to create storage during setup phase
           - name: RABBIT_NODE
             value: "0"
+          - name: NNF_EC_COMMAND_TIMEOUT
+            value: "90s"
           - name: NNF_NODE_NAME
             valueFrom:
               fieldRef:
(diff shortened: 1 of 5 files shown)


config/manager/manager.yaml
```diff
@@ -44,6 +44,8 @@ spec:
           # use this node to create storage during setup phase
           - name: RABBIT_NODE
             value: "0"
+          - name: NNF_EC_COMMAND_TIMEOUT
+            value: "90s"
```
bdevcich-hpe, line 48: For consistency with the k8s yaml below (i.e. `initialDelaySeconds` and in other places in rabbit), it might be better to name this `NNF_EC_COMMAND_TIMEOUT_SECONDS` and drop the `s` from `90s`.

matthew-richerson: That's what I had originally and I then I switched it to use time.ParseDuration() which can take different units ("ms", "s", "m", "h", etc.). If you think consistency with the other fields is more important I can switch it back. Seconds is the most realistic unit for this timeout.

bdevcich-hpe: The duraation strings definitely handy, but if seconds makes the most sense here then I think consistency should win.

go.mod
```diff
@@ -5,7 +5,7 @@ go 1.19
 require (
 	github.com/HewlettPackard/dws v0.0.0-20230410193930-8fd6fbeff7f8
 	github.com/NearNodeFlash/lustre-fs-operator v0.0.0-20230322153812-315a443f315b
-	github.com/NearNodeFlash/nnf-ec v0.0.0-20230425135615-c60e6e460894
+	github.com/NearNodeFlash/nnf-ec v0.0.0-20230426152614-be78544c8f4a
```
roehrich-hpe, line 8: Let's be careful about pointing into branches that are going to be deleted.  That can cause problems for someone who's trying to vendor on their own laptop, where didn't previously cache that branch.

roehrich-hpe (approved):

go.mod
matthew-richerson: I always re-vendor before merging to point at the commit in master. I just put out all the pull requests at once so you can see all the parts.

PR #195, opened 2023-05-16
Containers: Ensure mpiuser is mapped for MPI worker pods

bdevcich-hpe:
For MPI Container workflows, we need to ensure the `mpiuser` is properly mapped to the UID in the worker containers. The worker containers now have an InitContainer maps the user to the UID (like the Launcher container).

This ensures that the workflow's UID matches the permissions on the mounted filesystems inside of the worker pods, allowing `mpiuser` to perform IO.

The addition of the InitContainers to the worker nodes has caused a delay in the network accessibility from the Launcher pod to the Worker pods. To solve this, additional InitContainers have been added to the Launcher pod to make sure the Worker pods are resolvable via DNS before starting the Launcher.

That way, mpirun can properly hit the worker nodes.

This requires `dnsutils` is installed in the nnf-mfu container image to provide `nslookup`.

### DIFF
--- a/controllers/nnf_workflow_controller_helpers.go
+++ b/controllers/nnf_workflow_controller_helpers.go
@@ -1105,7 +1105,7 @@ func (r *NnfWorkflowReconciler) containerHandler(ctx context.Context, workflow *
 		})
 	}
 
-	addInitContainer := func(spec *corev1.PodSpec, user string, uid, gid int64, image string) {
+	addInitContainerPasswd := func(spec *corev1.PodSpec, user string, uid, gid int64, image string) {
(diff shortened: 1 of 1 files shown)


matthew-richerson (approved):

roehrich-hpe:
Func containerHandler() is now over 400 lines.  Please consider following this with a PR to break it up.  There are lots of opportunities to get lost while reading from the top to the bottom.

roehrich-hpe (approved):

bdevcich-hpe:
> Func containerHandler() is now over 400 lines. Please consider following this with a PR to break it up. There are lots of opportunities to get lost while reading from the top to the bottom.

I'm with you there.

PR #198, opened 2023-06-02
Use sh rather than bash in init containers

roehrich-hpe:
The alpine:latest container does not have a 'bash' executable.

Add some logging to indicate when the MPIJob or Job has been created.

### DIFF
--- a/controllers/nnf_workflow_controller_helpers.go
+++ b/controllers/nnf_workflow_controller_helpers.go
@@ -1392,7 +1392,7 @@ exit 1
 				return err
 			}
 		} else {
-			log.Info("Created MPI job", "name", mpiJob.Name, "namespace", mpiJob.Namespace)
+			log.Info("Created MPIJob", "name", mpiJob.Name, "namespace", mpiJob.Namespace)
 		}
 
 		return nil
--- a/controllers/nnf_workflow_controller_helpers.go
+++ b/controllers/nnf_workflow_controller_helpers.go
@@ -1051,7 +1051,7 @@ func (r *NnfWorkflowReconciler) removeAllPersistentStorageReferences(ctx context
 	return nil
 }
 
-func (r *NnfWorkflowReconciler) containerHandler(ctx context.Context, workflow *dwsv1alpha2.Workflow, dwArgs map[string]string, index int) (*result, error) {
+func (r *NnfWorkflowReconciler) containerHandler(ctx context.Context, workflow *dwsv1alpha2.Workflow, dwArgs map[string]string, index int, log logr.Logger) (*result, error) {
 	profile, err := r.getContainerProfile(ctx, workflow, index)
 	if err != nil {
 		return nil, err
@@ -1127,7 +1127,7 @@ exit 0
 			Name:  "mpi-init-passwd",
 			Image: image,
 			Command: []string{
-				"/bin/bash",
+				"/bin/sh",
 				"-c",
 				script,
 			},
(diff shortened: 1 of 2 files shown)


ajfloeder (approved):

controllers/nnf_workflow_controller_helpers.go
```diff
@@ -1391,6 +1391,8 @@ exit 1
 			if !apierrors.IsAlreadyExists(err) {
 				return err
 			}
+		} else {
+			log.Info("Created MPI job", "name", mpiJob.Name, "namespace", mpiJob.Namespace)
```
bdevcich-hpe, line 1395: Might want to say MPIJob here since that's the resource type.

controllers/nnf_workflow_controller_helpers.go
```diff
@@ -1164,7 +1164,7 @@ exit 1
 			Name:  fmt.Sprintf("mpi-wait-for-worker-%d", worker),
 			Image: spec.Containers[0].Image,
 			Command: []string{
-				"/bin/bash",
+				"/bin/sh",
```
bdevcich-hpe, line 1167: This actually breaks the script because using `sh` here rather than `bash` causes the script to bomb out when `nslookup` fails instead of looping.

I guess I'm still confused how using bash caused a failure for you here.

roehrich-hpe: The container alpine:latest, when pulled for my M1 mac, does not have a 'bash' executable anywhere.

```
: dean-mbp; docker run -it --rm --name c1 alpine:latest sh               
Unable to find image 'alpine:latest' locally
latest: Pulling from library/alpine
08409d417260: Pull complete 
Digest: sha256:02bb6f428431fbc2809c5d1b41eab5a68350194fb508869a33cb1af4444c9b11
Status: Downloaded newer image for alpine:latest
/ # ls
bin    etc    lib    mnt    proc   run    srv    tmp    var
dev    home   media  opt    root   sbin   sys    usr
/ # type bash
bash: not found
/ # ls /bin/bash
ls: /bin/bash: No such file or directory
/ # type sh
sh is /bin/sh
/ # ls /usr/bin/bash
ls: /usr/bin/bash: No such file or directory
/ # exit
```

roehrich-hpe: So that for-loop doesn't work in 'sh':
```
: dean-mbp; docker run -it --rm --name c1 alpine:latest sh
/ # for i in {1..5}; do echo $i; done
{1..5}
```


I'll put up another PR to switch it to use `seq`:
```
/ # for i in $(seq 1 5); do echo $i; done
1
2
3
4
5
```

roehrich-hpe: I see this same lack of bash when I pull it to dp1a:
```
[root@rabbit-k8s-master ~]# docker run -it --rm --name c1 alpine:latest sh
Unable to find image 'alpine:latest' locally
latest: Pulling from library/alpine
8a49fdb3b6a5: Pull complete 
Digest: sha256:02bb6f428431fbc2809c5d1b41eab5a68350194fb508869a33cb1af4444c9b11
Status: Downloaded newer image for alpine:latest
/ # type bash
bash: not found
/ # ls /bin/bash
ls: /bin/bash: No such file or directory
/ # 
```

roehrich-hpe: So it's not an issue with the M1 builds.

roehrich-hpe: PR posted in https://github.com/NearNodeFlash/nnf-sos/pull/201

PR #203, opened 2023-06-13
Update controller-tools to v0.12.0

roehrich-hpe:
The older controller-tools would put a creationTimestamp in the resources and would give it a "null" value.

Re-run "make manifests" to let the new controller-gen cleanup this stuff.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,8 @@
 module github.com/NearNodeFlash/nnf-sos
 
+replace github.com/NearNodeFlash/lustre-fs-operator => ../lustre-fs-operator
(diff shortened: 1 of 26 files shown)


go.mod
```diff
@@ -1,5 +1,8 @@
 module github.com/NearNodeFlash/nnf-sos
 
+replace github.com/NearNodeFlash/lustre-fs-operator => ../lustre-fs-operator
```
ajfloeder, line 3: Don't merge this

bdevcich-hpe (approved):

PR #207, opened 2023-06-15
Refactored containerHandler()

bdevcich-hpe:
`containerHandler()` was much too large. This change moves all of the underlying User Container code into it's own helper file. To reduce the number of function parameters (e.g. workflow, profile, ctx) passed into these functions, a new struct was added so that the helper functions could make use it.

Also renamed it to `userContainerHandler()`.

### DIFF
--- a/controllers/nnf_workflow_controller_container_helpers.go
+++ b/controllers/nnf_workflow_controller_container_helpers.go
@@ -27,13 +27,16 @@ import (
 
 	dwsv1alpha2 "github.com/HewlettPackard/dws/api/v1alpha2"
 	nnfv1alpha1 "github.com/NearNodeFlash/nnf-sos/api/v1alpha1"
+	"github.com/go-logr/logr"
 	mpicommonv1 "github.com/kubeflow/common/pkg/apis/common/v1"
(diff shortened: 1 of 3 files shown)


controllers/nnf_workflow_controller_container_helpers.go
```diff
@@ -0,0 +1,434 @@
...
+)
+
+type nnfUserContainer struct {
+	workflow *dwsv1alpha2.Workflow
+	profile  *nnfv1alpha1.NnfContainerProfile
+	nnfNodes []string
+	volumes  []nnfContainerVolume
+	username string
+	uid, gid int64
+	r        *NnfWorkflowReconciler
```
bdevcich-hpe, line 46: Not sure how I feel about this, but we do need it for creating resources.

roehrich-hpe, line 46: Send in the parts of 'r' that you need.  So use these two fields, instead:
```
 clnt client.Client
log logr.Logger
```

Then set those fields with r.Client and r.Log.

bdevcich-hpe: Perfect. Thanks!

roehrich-hpe (approved):

PR #208, opened 2023-06-20
Allow NnfContainerProfiles to be restricted to a UserID/GroupID

roehrich-hpe:
Add an optional UserID/GroupID field to NnfContainerProfileData.  If either of these is specified, then the chosen container profile must match the user ID or group ID specified in the workflow resource that references the profile.

Add tests.

### DIFF
--- a/controllers/nnf_workflow_controller_test.go
+++ b/controllers/nnf_workflow_controller_test.go
@@ -47,16 +47,26 @@ import (
 // BeforeEach - initialize the workflow
 // AfterEach - destroy the workflow
 
+var (
+	baseWorkflowUserID  uint32 = 1042
+	baseWorkflowGroupID uint32 = 1043
+
+	altWorkflowUserID  uint32 = 1044
+	altWorkflowGroupID uint32 = 1045
+)
+
 var _ = Describe("NNF Workflow Unit Tests", func() {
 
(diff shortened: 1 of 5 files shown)


controllers/nnf_workflow_controller_test.go
```diff
@@ -1079,29 +1077,68 @@ var _ = Describe("NNF Workflow Unit Tests", func() {
...
+			)
+
+			DescribeTable("should not go to Proposal Ready when profile restriction is not satisfied",
+				func(containerUserID *uint32, containerGroupID *uint32) {
+					buildRestrictedContainerProfile(containerUserID, containerGroupID)
+					buildWorkflowWithCorrectDirectives()
+					Eventually(func(g Gomega) bool {
+						g.Expect(k8sClient.Get(context.TODO(), key, workflow)).To(Succeed())
+						return workflow.Status.Status == dwsv1alpha2.StatusError && strings.Contains(workflow.Status.Message, "container profile") && strings.Contains(workflow.Status.Message, "is restricted to")
+					}).Should(BeFalse(), "does not reach desired Proposal state")
```
roehrich-hpe, line 1136: That's wrong...

bdevcich-hpe (approved):

PR #209, opened 2023-06-22
Containers: Add specific linux capabilities for MPI workers

bdevcich-hpe:
On LLNL systems, the host's default capabilities are different than HPE systems due to the difference in container runtime and also configuration of those runtimes.

This change drops all capabilities and then adds in the required ones to ensure mpirun can access the worker pods.

Additionally, fix some bugs in the MPISpec of the NnfContainerProfile:
- Don't allow NnfContainerProfile Data to be empty, it's required
- Added check to ensure MPISpec is defined
- Added checks to ensure launcher/worker are defined in the MPISpec

Tested capabilities by:
- Drop ALL capabilities without the specific Adds (and failing)
- Then adding in the specific Adds (and passing)

### DIFF
--- a/controllers/nnf_workflow_controller_container_helpers.go
+++ b/controllers/nnf_workflow_controller_container_helpers.go
@@ -73,6 +73,11 @@ func (c *nnfUserContainer) createMPIJob() error {
 			Namespace: c.workflow.Namespace,
 		},
 	}
+
+	if c.profile.Data.MPISpec == nil {
+		return fmt.Errorf("Profile does not contain Data.MPISpec")
+	}
+
 	c.profile.Data.MPISpec.DeepCopyInto(&mpiJob.Spec)
 	c.username = nnfv1alpha1.ContainerMPIUser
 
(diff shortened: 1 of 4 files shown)


controllers/nnf_workflow_controller_container_helpers.go
```diff
@@ -73,6 +73,11 @@ func (c *nnfUserContainer) createMPIJob() error {
 			Namespace: c.workflow.Namespace,
 		},
 	}
+
+	if c.profile.Data.MPISpec == nil {
```
roehrich-hpe, line 77: This, and the MPIReplicaSpecs checks below, should be moved to the webhook...in fact, the webhook already has this stuff.

bdevcich-hpe: This is stuff I ran into when attempting to unit test (i.e. no webhook). I suppose this is a belt and suspenders approach. Too much?

controllers/nnf_workflow_controller_container_helpers.go
```diff
@@ -86,9 +91,16 @@ func (c *nnfUserContainer) createMPIJob() error {
 	}
 
 	// MPIJobs have two pod specs: one for the launcher and one for the workers
-	launcher := mpiJob.Spec.MPIReplicaSpecs[mpiv2beta1.MPIReplicaTypeLauncher]
+	launcher, found := mpiJob.Spec.MPIReplicaSpecs[mpiv2beta1.MPIReplicaTypeLauncher]
+	if !found {
+		return fmt.Errorf("%s not found in MPIReplicaSpecs", mpiv2beta1.MPIReplicaTypeLauncher)
```
ajfloeder, line 96: Would this allow an admin to root cause the issue?
Same question on line 102 below.

controllers/nnf_workflow_controller_container_helpers.go
roehrich-hpe: I would have expected all of that to be caught here: 
https://github.com/NearNodeFlash/nnf-sos/blob/master/api/v1alpha1/nnfcontainerprofile_webhook.go#L51-L80
Does something in the webhook need to be adjusted?

roehrich-hpe: How were you unit-testing this without having the webhook running?

roehrich-hpe: Well this is fun:

In suite_test.go, comment out the webhook:
https://github.com/NearNodeFlash/nnf-sos/blob/master/controllers/suite_test.go#L184

Then notice that the webhook-present test doesn't fail:
https://github.com/NearNodeFlash/nnf-sos/blob/master/controllers/nnf_workflow_controller_test.go#L1241

bdevcich-hpe: > How were you unit-testing this without having the webhook running?

I was able to get this method unit tested up until the `Create()` point - where mpi-operator is needed. Seems like I was only able to get that far based on your findings that the webhook won't fail? I'll dig into this more.

bdevcich-hpe: I've removed the checks in favor of the webhook.

controllers/nnf_workflow_controller_container_helpers.go
bdevcich-hpe: Per Dean's comments, I've removed this checks in favor of the webhook. The admin will get a helpful error message upon NnfContainerProfile creation due to the webhook if the launcher/worker specs are missing.

roehrich-hpe (approved):

PR #210, opened 2023-06-23
Tighten the tests that verify the profile webhooks are hooked up.

roehrich-hpe:
The NnfContainerProfile test was targeting a failure in the kube-apiserver, by tripping a kubebuilder:validation for RetryLimit.  That has been changed to something that will pass the kube-apiserver, and then will proceed on to fail in the webhook.

Now we can distinguish when Create() errors are responses from the webhook, versus when they are errors about the webhook not being installed.


ajfloeder (approved):

bdevcich-hpe (approved):
Thanks for fixing this.

PR #213, opened 2023-07-10
Containers: Add global lustre support

bdevcich-hpe:
This adds `DW_GLOBAL_*` storage support to container workflows/profiles, which are backed by the `LustreFilesystem` resource. This allows users to mount global lustre fileystems into user containers.

An administrator must first edit the `LustreFilesystem` resource to add the workflow's namespace (e.g. `default`) to the list of namespaces along with a mode. The mode defaults to `ReadWriteMany` if not set. Doing so will create a PVC that can be used to mount the fileystem.

A `DW_GLOBAL_*` storage must also be added to the NnfContainer Profile. See `config/examples` and `config/samples` for more detail.

The value for the `DW_GLOBAL_*` parameter is the path of the lustre filesystem, e.g.:

```
#DW jobdw type=gfs2 name=my-local-storage capacity=100GB
#DW container name=my-container-workflow profile=example-mpi \
    DW_JOB_foo_local_storage=my-local-storage \
    DW_GLOBAL_foo_global_lustre=/lus/sawbill
```

### DIFF
--- a/api/v1alpha1/nnfcontainerprofile_webhook_test.go
+++ b/api/v1alpha1/nnfcontainerprofile_webhook_test.go
@@ -61,6 +61,11 @@ var _ = Describe("NnfContainerProfile Webhook", func() {
 						{Name: "test"},
 					},
 				},
+				Storages: []NnfContainerProfileStorage{
+					{Name: "DW_JOB_storage", Optional: true},
+					{Name: "DW_PERSISTENT_storage", Optional: true},
+					{Name: "DW_GLOBAL_storage", Optional: true},
+				},
 			},
 		}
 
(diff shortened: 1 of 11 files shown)


api/v1alpha1/nnfcontainerprofile_webhook_test.go
```diff
@@ -304,4 +309,28 @@ var _ = Describe("NnfContainerProfile Webhook", func() {
...
+			}
+			if result {
+				Expect(k8sClient.Create(context.TODO(), nnfProfile)).To(Succeed())
+			} else {
+				Expect(k8sClient.Create(context.TODO(), nnfProfile)).ToNot(Succeed())
+				nnfProfile = nil
+			}
+		},
+		// Only nil modes should pass for JOB/PERSISTENT
+		Entry("should pass when DW_JOB has no mode", "DW_JOB_storage", nil, true),
```
roehrich-hpe, line 328: PersistentVolumeAccessMode is a string, not a string pointer.  You could redefine 'mode' as `*corev1.PersistentVolumeAccessmode`, and adjust the conditional for that.  Or you could send an empty string.

I think it's confusing when you send a nil for something that's not a pointer.

roehrich-hpe (approved):

matthew-richerson (approved):

PR #214, opened 2023-07-25
Containers: Allow containers to open ports for communication with com…

bdevcich-hpe:
…pute nodes

This change allows for container workflows to open ports. These ports are opened on the host nodes (i.e. NNF nodes) where the containers are running. This enables traffic from outside of the network through the IP address of the NNF node and the port. An application on the compute node can contact the container with <NNF_NODE_IP>:<PORT>.

The port number(s) can be retrieved via the NNF_CONTAINER_PORT environment variable. This environment variable is available inside of the containers. It is also provided to the Workflow so that Flux can inform the application on the compute node of which port(s) to use. If multiple ports are desired, the environment variable will provide a comma separated list of port numbers.

Ports are requested via the NnfContainerProfile's `numPorts`. **A system admin must enable the `Ports` port range in the `SystemConfiguration` before ports can be requested**. If not, the NnfPortManager will not allocate any ports.

More details:
- Enabled default NnfPortManager to manage port allocation
- Port allocation occurs in the Setup State
- Port de-allocation occurs in the Teardown State
- User Container Pods are now destroyed in the Teardown State prior to Port de-allocation
- Added `example-mpi-webserver` NnfContainerProfile to show use of envionrment variable with a simple webserver
- Added container teardown + port allocation to workflow deletion


controllers/nnf_workflow_controller_helpers.go
```diff
@@ -1547,3 +1625,146 @@ func (r *NnfWorkflowReconciler) getContainerVolumes(ctx context.Context, workflo
...
+
+	return nil, nil
+}
+
+// Retrieve the default NnfPortManager for user containers. Allow a client to be passed in as this
+// is meant to be used by reconcilers or container helpers.
+func getContainerPortManager(ctx context.Context, cl client.Client) (*nnfv1alpha1.NnfPortManager, error) {
+	pm := &nnfv1alpha1.NnfPortManager{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "nnf-port-manager",
```
roehrich-hpe, line 1717: This name and namespace should be sent in via env vars:
https://github.com/NearNodeFlash/nnf-sos/blob/master/config/manager/manager.yaml#L110
and
https://github.com/NearNodeFlash/nnf-sos/blob/master/controllers/suite_test.go#L74

controllers/nnf_workflow_controller_helpers.go
```diff
@@ -1246,6 +1246,66 @@ func (r *NnfWorkflowReconciler) waitForContainersToStart(ctx context.Context, wo
...
+	// Add workflow matchLabels + directive index (if desired)
+	matchLabels := dwsv1alpha2.MatchingWorkflow(workflow)
+	if index >= 0 {
+		matchLabels[nnfv1alpha1.DirectiveIndexLabel] = strconv.Itoa(index)
+	}
+
+	// Delete MPIJobs
+	mpiJobList, err := r.getMPIJobs(ctx, workflow, index)
+	if err != nil {
+		if strings.Contains(err.Error(), "no kind is registered for the type v2beta1.MPIJobList") || apierrors.IsNotFound(err) {
```
roehrich-hpe, line 1269: I would trim this string to "no kind is registered for the type", just to future-proof it a bit.

controllers/nnf_workflow_controller_helpers.go
```diff
@@ -1246,6 +1246,66 @@ func (r *NnfWorkflowReconciler) waitForContainersToStart(ctx context.Context, wo
...
+			doneMpi = true
+		} else {
+			return nil, nnfv1alpha1.NewWorkflowError("Could not delete container MPIJob(s)").WithError(err)
+		}
+	} else if len(mpiJobList.Items) > 0 {
+		if err := r.DeleteAllOf(ctx, &mpiJobList.Items[0], client.InNamespace(workflow.Namespace), matchLabels, deleteAllOptions); err != nil {
+			if !apierrors.IsNotFound(err) {
+				return nil, nnfv1alpha1.NewWorkflowError("Could not delete container MPIJob(s)").WithError(err)
+			}
```
roehrich-hpe, line 1278: Have you seen any other kind of error here?
If you know we can ignore any other errors here then say that in a comment.

controllers/nnf_workflow_controller_helpers.go
```diff
@@ -1442,14 +1503,31 @@ func (r *NnfWorkflowReconciler) getMPIJobList(ctx context.Context, workflow *dws
...
+	}
+
+	return jobList, nil
+}
+
 func (r *NnfWorkflowReconciler) getContainerJobs(ctx context.Context, workflow *dwsv1alpha2.Workflow, index int) (*batchv1.JobList, error) {
 	// Get the jobs for this workflow and directive index
 	matchLabels := dwsv1alpha2.MatchingWorkflow(workflow)
-	matchLabels[nnfv1alpha1.DirectiveIndexLabel] = strconv.Itoa(index)
+	if index >= 0 {
```
roehrich-hpe, line 1524: What would cause it to be less than zero?  (I'm still digging...)

bdevcich-hpe: I made this and `getMPIJobs` have the ability to use `-1` for `index` for when you don't want to filter by index. `deleteContainers` is called with an index of `-1` when the workflow is deleted (rather than a proper Teardown->delete). See here: https://github.com/NearNodeFlash/nnf-sos/blob/da071ea41ed1194601ca43901f9dbb474837736d/controllers/nnf_workflow_controller.go#L128

controllers/nnf_workflow_controller_helpers.go
```diff
@@ -1547,3 +1625,146 @@ func (r *NnfWorkflowReconciler) getContainerVolumes(ctx context.Context, workflo
...
+	// Nothing to do here if ports are not requested
+	if profile.Data.NumPorts > 0 {
+		pm, err := getContainerPortManager(ctx, r.Client)
+		if err != nil {
+			return nil, err
+		}
+
+		// Check to see if we've already made an allocation
+		for _, alloc := range pm.Spec.Allocations {
+			if alloc.Requester.UID == workflow.UID {
```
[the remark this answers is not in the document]
bdevcich-hpe: Correct: https://github.com/NearNodeFlash/nnf-sos/blob/3d439e4f39d325bfe4cf186028da22152148e476/controllers/nnf_workflow_controller_helpers.go#L148-L151

controllers/nnf_workflow_controller_container_helpers.go
```diff
@@ -391,6 +409,86 @@ func (c *nnfUserContainer) applyPermissions(spec *corev1.PodSpec, mpiJobSpec *mp
...
+}
+
+// Given a list of ports, convert it into an environment variable name and comma separated value
+func getContainerPortsEnvVar(ports []uint16) (string, string) {
+	portStr := []string{}
+	for _, port := range ports {
+		portStr = append(portStr, strconv.Itoa(int(port)))
+	}
+
+	return "NNF_CONTAINER_PORT", strings.Join(portStr, ",")
```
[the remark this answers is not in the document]
bdevcich-hpe: I can change it.

controllers/nnf_workflow_controller_helpers.go
bdevcich-hpe: I haven't seen other errors here.

controllers/nnf_workflow_controller.go
```diff
@@ -412,38 +426,48 @@ func (r *NnfWorkflowReconciler) startSetupState(ctx context.Context, workflow *d
...
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      name,
-			Namespace: namespace,
-		},
-	}
-	if err := r.Get(ctx, client.ObjectKeyFromObject(nnfStorage), nnfStorage); err != nil {
-		return nil, nnfv1alpha1.NewWorkflowErrorf("failed to get NNF storage resource '%s", client.ObjectKeyFromObject(nnfStorage)).WithError(err)
-	}
+	switch dwArgs["command"] {
+	default:
```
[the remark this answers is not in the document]
bdevcich-hpe: I'll change it around

controllers/nnf_workflow_controller_container_helpers.go
```diff
@@ -391,6 +409,86 @@ func (c *nnfUserContainer) applyPermissions(spec *corev1.PodSpec, mpiJobSpec *mp
...
+	}
+
+	pm, err := getContainerPortManager(c.ctx, c.client)
+	if err != nil {
+		return nil, err
+	}
+
+	// Get the ports from the port manager for this workflow
+	for _, alloc := range pm.Status.Allocations {
+		if alloc.Requester != nil && alloc.Requester.UID == c.workflow.UID {
```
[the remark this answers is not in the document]
bdevcich-hpe: Yes, good call.

controllers/nnf_workflow_controller.go
```diff
@@ -974,6 +998,14 @@ func (r *NnfWorkflowReconciler) finishPostRunState(ctx context.Context, workflow
 }
 
 func (r *NnfWorkflowReconciler) startTeardownState(ctx context.Context, workflow *dwsv1alpha2.Workflow, index int) (*result, error) {
+	dwArgs, _ := dwdparse.BuildArgsMap(workflow.Spec.DWDirectives[index])
```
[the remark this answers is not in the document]
bdevcich-hpe: I'll change it to a switch

controllers/nnf_workflow_controller_helpers.go
bdevcich-hpe: Done

roehrich-hpe (approved):

bdevcich-hpe:
I'm waiting on #215 and then I will rebase and refactor all the new error paths.

bdevcich-hpe:
I've rebased on top of the new error reporting and accounted for the new errors. Please take another look, especially the latest commit.

controllers/nnf_workflow_controller_container_helpers.go
```diff
@@ -431,10 +431,10 @@ func (c *nnfUserContainer) getHostPorts() ([]uint16, error) {
 
 	// Make sure we found the number of ports in the port manager that we expect
 	if len(ports) != expectedPorts {
-		return nil, nnfv1alpha1.NewWorkflowError(fmt.Sprintf(
+		return nil, dwsv1alpha2.NewResourceError(
 			"number of ports found in NnfPortManager's allocation (%d) does not equal the profile's requested ports (%d)",
-			len(ports), expectedPorts)).
-			WithFatal()
+			len(ports), expectedPorts).
+			WithUserMessage("requested ports do not meet the number of allocated ports").WithFatal().WithInternal()
```
matthew-richerson, line 437: WithInternal() is the default when using NewResourceError(), so it's not necessary to add it. It doesn't hurt anything though

matthew-richerson (approved):

controllers/nnf_workflow_controller_container_helpers.go
bdevcich-hpe: I'll take that out.

config/default/kustomization.yaml
```diff
@@ -19,6 +19,7 @@ bases:
 # [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix including the one in
 # crd/kustomization.yaml
 - ../webhook
+- ../ports
```
bdevcich-hpe, line 22: Apparently even with this change, it's still possible to run into this error during deployment: 

```
error: resource mapping not found for name: "nnf-port-manager" namespace: "nnf-system" from "STDIN": no matches for kind "NnfPortManager" in version "nnf.cray.hpe.com/v1alpha1"
ensure CRDs are installed first
```

roehrich-hpe: Yes.  It has to be its own 'kubectl apply', so you can put a wait/watch between them to check for the CRD being ready.

bdevcich-hpe: Thanks @roehrich-hpe. I started down this path and it seems to be working. Update incoming.

bdevcich-hpe: Fixed.

deploy.sh
```diff
@@ -52,9 +52,18 @@ if [[ $CMD == 'deploy' ]]; then
     # MPIJobSpec (with large annotations).
     (cd config/examples && $KUSTOMIZE edit set image nnf-mfu=$NNFMFU_IMG)
     $KUSTOMIZE build config/examples | kubectl apply --server-side=true --force-conflicts -f -
+
+    # Deploy the nnfportmanager after everything else
+    echo "Waiting for the nnfportmamanger CRD to become ready..."
+    while :; do
+        kubectl get crds nnfportmanagers.nnf.cray.hpe.com && break
+        sleep 1
```
roehrich-hpe, line 60: Might want to lead with the sleep.

bdevcich-hpe: done

PR #215, opened 2023-07-28
github-46: Use ResourceError when returning errors

matthew-richerson:
This commit uses the new ResourceError struct embedded in the status section of the DWS/nnf-sos resources. When returning an error, use the NewResourceError() call to return a ResourceError and fill it in with the correct information. This allows the end user and WLM to make informed decisions about what to do when there's an error.


bdevcich-hpe (approved):

bdevcich-hpe:
Nice work!

controllers/dws_servers_controller.go
```diff
@@ -21,7 +21,6 @@ package controllers
```
roehrich-hpe, line 21: Update the copyright date on this file and each of the others.

controllers/nnf_workflow_controller_test.go
```diff
@@ -167,6 +167,14 @@ var _ = Describe("NNF Workflow Unit Tests", func() {
 				FileSystemType: fsType,
 				AllocationSets: []nnfv1alpha1.NnfStorageAllocationSetSpec{},
 			},
+			Status: nnfv1alpha1.NnfStorageStatus{
```
roehrich-hpe, line 170: NnfStorage defines Status as a subresource, so the `k8sClient.Create()` below will not store this status section.

roehrich-hpe (approved):

PR #238, opened 2023-11-14
Check Data Movement Manager for Ready Flag for DataIn/DataOut

bdevcich:
- Add check in DataIn/DataOut and requeue when DMM is not ready
- Remove omitempty from ddm ready flag so this shows up as false when not
ready.
- RBAC fix


internal/controller/integration_test.go
```diff
@@ -966,20 +967,45 @@ var _ = Describe("Integration Test", func() {
...
+			dmm = &nnfv1alpha1.NnfDataMovementManager{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      nnfv1alpha1.DataMovementManagerName,
+					Namespace: nnfv1alpha1.DataMovementNamespace,
+				},
+				Spec: nnfv1alpha1.NnfDataMovementManagerSpec{
+					Template: corev1.PodTemplateSpec{
+						Spec: corev1.PodSpec{
+							Containers: []corev1.Container{{
+								Name:  "name-manager dummy",
```
roehrich-hpe, line 980: That doesn't look like a valid value for a k8s resource name.

bdevcich: This is never actually used, but I fixed it.

ajfloeder (approved):

internal/controller/nnf_workflow_controller.go
```diff
@@ -608,6 +608,18 @@ func (r *NnfWorkflowReconciler) startDataInOutState(ctx context.Context, workflo
...
+	// Verify data movement is ready
+	dmm := &nnfv1alpha1.NnfDataMovementManager{ObjectMeta: metav1.ObjectMeta{
+		Name:      nnfv1alpha1.DataMovementManagerName,
+		Namespace: nnfv1alpha1.DataMovementNamespace,
+	}}
+	if err := r.Get(ctx, client.ObjectKeyFromObject(dmm), dmm); err != nil {
+		return nil, dwsv1alpha2.NewResourceError("could not get NnfDataMovementManager %v", client.ObjectKeyFromObject(dmm)).WithError(err).WithUserMessage("could not determine data movement readiness")
+	}
+	if !dmm.Status.Ready {
+		return Requeue("pending data movement readiness").withObject(dmm), nil
```
matthew-richerson, line 620: I think this would be better requeuing after a time delay

internal/controller/integration_test.go
bdevcich: Just to clarify - this container spec is never used because the DMM reconciler doesn't actually run, but we do need to specify certain fields (name, image) to make the podspec happy.

matthew-richerson (approved):

PR #242, opened 2023-12-15
File system and block device management in nnf-sos

matthew-richerson:
This commit moves the file system, LVM, and zpool management out of nnf-ec and into nnf-sos. There are many parts to this change:

- Create a package for running the LVM commands (e.g., pvcreate)
- Create a package for managing block devices. This presents as the blockdevice interface. Implementations include lvm, zpool, and mock.
- Create a package for managing file systems. This presents as the filesystem interface. Implementations include simple (for xfs, gfs2 , and bind mounts), lustre, mock, and kind.
- Create an NnfNodeBlockStorage resource that interfaces with nnf-ec to manage the storage pools and storage groups.
- Change NnfNodeStorage to use the blockdevice and filesystem interfaces to create and delete file systems on the Rabbit.
- Change the ClientMount reconciler to also use the blockdevice and filesystem interfaces to activate/mount and unmount/deactivate filesystems and blockdevices.
- Move the clientmount daemon from DWS to nnf-sos. The ClientMount reconciler is now shared between compute and Rabbit nodes (run as clientmountd on the computes and inside nnf-node-manager on the Rabbits), and it is NNF specific (not generic for DWS).
- Add command line arguments to the NnfStorageProfile for activating/deactivating single LVs
- Use a shared VG with locking for all LVM VGs since they're always shared between multiple nodes. This causes LVM to enforce a single node activating a non-shared LV at a time. In the future we can remove the internal locking nnf-sos does in the NnfAccess reconciler
- Use the ClientMount resource to mount the Lustre file system after creation to set the GID/UID
- Try to consistently use MgsAddress as the resource field for the MGS LNet Address
- Turn NnfNodeStorage Status section into a subresource


config/prometheus/kustomization.yaml
```diff
@@ -1,5 +1,5 @@
 # Adds namespace to all resources.
 namespace: nnf-system
 
-resources:
-- monitor.yaml
+#resources:
```
roehrich-hpe, line 4: Turn this back on, please.

matthew-richerson: Oops. Yep, I'll put that back

api/v1alpha1/nnf_node_block_storage_types.go
```diff
@@ -0,0 +1,127 @@
+/*
+ * Copyright 2021-2023 Hewlett Packard Enterprise Development LP
```
bdevcich, line 2: Just 2023 here

api/v1alpha1/nnf_node_block_storage_types.go
```diff
@@ -0,0 +1,127 @@
...
+type NnfNodeBlockStorageAllocationSpec struct {
+	// Aggregate capacity of the block devices for each allocation
+	Capacity int64 `json:"capacity,omitempty"`
+
+	// List of nodes where /dev devices should be created
+	Access []string `json:"access,omitempty"`
+}
+
+// NnfNodeBlockStorageSpec defines the desired storage attributes on a NNF Node.
+// Storage spec are created on bequest of the user and fullfilled by the NNF Node Controller.
```
bdevcich, line 38: request?

api/v1alpha1/nnf_node_block_storage_types.go
```diff
@@ -0,0 +1,127 @@
...
+type NnfNodeBlockStorageSpec struct {
+	// Allocations is the list of storage allocation to make
```
bdevcich, line 40:
```suggestion
	// Allocations is the list of storage allocations to make
```

api/v1alpha1/nnf_node_block_storage_types.go
```diff
@@ -0,0 +1,127 @@
...
+
+type NnfNodeBlockStorageDeviceStatus struct {
+	// NQN of the base NVMe device
+	NQN string `json:"NQN"`
+
+	// Id of the Namespace on the NVMe device (e.g., "2")
+	NamespaceId string `json:"namespaceId"`
+
+	// Total capacity allocated for the storage. This may differ from the requested storage
+	// capacity as the system may round up to the requested capacity satisify underlying
```
bdevcich, line 61: satisfying?

api/v1alpha1/nnf_node_block_storage_types.go
```diff
@@ -0,0 +1,127 @@
...
+type NnfNodeBlockStorageAccessStatus struct {
+	// /dev paths for each of the block devices
+	DevicePaths []string `json:"devicePaths,omitempty"`
+
+	// Redfish ID for the storage group
+	StorageGroupId string `json:"storageGroupId,omitempty"`
+}
+
+type NnfNodeBlockStorageAllocationStatus struct {
+	Accesses map[string]NnfNodeBlockStorageAccessStatus `json:"accesses,omitempty"`
```
bdevcich, line 75: Accesses should probaly have a comment as well

api/v1alpha1/nnf_node_block_storage_types.go
```diff
@@ -0,0 +1,127 @@
...
+
+	// List of NVMe namespaces used by this allocation
+	Devices []NnfNodeBlockStorageDeviceStatus `json:"devices,omitempty"`
+
+	// Total capacity allocated for the storage. This may differ from the requested storage
+	// capacity as the system may round up to the requested capacity satisify underlying
```
bdevcich, line 81: stastify here too - satsifying or to satfisy?

matthew-richerson: These were all just copied from NnfNodeStorage. I'll fix the spelling

internal/controller/dws_servers_controller.go
```diff
@@ -195,7 +195,7 @@ func (r *DWSServersReconciler) updateCapacityUsed(ctx context.Context, servers *
 
 		allocationSet := nnfStorage.Status.AllocationSets[storageIndex]
 
-		if allocationSet.Status != nnfv1alpha1.ResourceReady {
+		if allocationSet.Ready == false {
 			ready = false
```
bdevcich, line 199: Couldn't you just set `false = allocation.Ready`?

internal/controller/nnf_clientmount_controller.go
```diff
@@ -192,138 +177,80 @@ func (r *NnfClientMountReconciler) changeMountAll(ctx context.Context, clientMou
...
+			Name:      clientMount.Spec.Mounts[index].Device.DeviceReference.ObjectReference.Name,
+			Namespace: clientMount.Spec.Mounts[index].Device.DeviceReference.ObjectReference.Namespace,
+		},
 	}
 
-	return fileShare, nil
-}
+	// These labels aren't exactly right (NnfStorage owns NnfNodeStorage), but the
+	// labels that are important for doing the mount are there and correct
+	dwsv1alpha2.InheritParentLabels(nnfNodeStorage, clientMount)
```
bdevcich, line 231: Should this be fixed?

internal/controller/nnf_storage_controller.go
```diff
@@ -291,13 +316,149 @@ func (r *NnfStorageReconciler) removePersistentStorageReference(ctx context.Cont
...
+		matchLabels,
+	}
+
+	if err := r.List(ctx, nnfNodeBlockStorageList, listOptions...); err != nil {
+		return &ctrl.Result{Requeue: true}, nil
+	}
+
+	// Ensure that we found all the NnfNodeStorage resources we were expecting
+	if len(nnfNodeBlockStorageList.Items) != len(nnfStorage.Spec.AllocationSets[allocationSetIndex].Nodes) {
+		return &ctrl.Result{}, nil
```
bdevcich, line 398: shouldn't this error or requeue?

bdevcich (approved):

api/v1alpha1/nnf_node_storage_types.go
```diff
@@ -83,7 +65,7 @@ type LustreStorageSpec struct {
 	FileSystemName string `json:"fileSystemName,omitempty"`
 
 	// TargetType is the type of Lustre target to be created.
-	// +kubebuilder:validation:Enum=MGT;MDT;MGTMDT;OST
+	// +kubebuilder:validation:Enum=mgt;mdt;mgtmdt;ost
```
roehrich-hpe, line 68: What is the reason for switching the case here?  I'm concerned about this causing breakage in other areas.  Does Flux need to know about this change?

Also, there are still uppercase references in `config/samples`.

api/v1alpha1/nnfstorageprofile_types.go
```diff
@@ -137,49 +140,47 @@ type NnfStorageProfileCmdLines struct {
...
 	LvCreate string `json:"lvCreate,omitempty"`
 
+	// LvChange specifies the various lvchange commandlines, minus the "lvchange"
+	LvChange NnfStorageProfileLVMLvChangeCmdLines `json:"lvChange,omitempty"`
+
 	// LvRemove specifies the lvcreate commandline, minus the "lvremove".
 	LvRemove string `json:"lvRemove,omitempty"`
+
+	// MountRabbit specifies mount options for mounting on the Rabbit.
+	MountRabbit string `json:"mountRabbit,omitempty"`
```
roehrich-hpe, line 150: Are we using MountRabbit and MountCompute anymore?

internal/controller/dws_servers_controller.go
matthew-richerson: Yep... will change

internal/controller/nnf_clientmount_controller.go
matthew-richerson: This is done on purpose. It's making a fake NnfNodeStorage resource instead of reading the actual NnfNodeStorage from the API server. If we wanted to read the actual NnfNodeStorage, we'd need the Rabbit's namespace added to the Clientmountd client cache. That would mean the watch()'s in Clientmountd would get events from the Rabbit namespace as well. That would put a lot more load on the API server and increase jitter on the compute. We have enough information contained in the ClientMount resource to populate the important fields in a fake NnfNodeStorage that only lives in memory for the Reconcile() call. A couple of the labels are wrong, but they're not used for anything so it doesn't matter.

internal/controller/nnf_storage_controller.go
matthew-richerson: All of the NnfNodeBlockStorages were created in createNodeBlockStorage(), so it they don't exist here yet it's because the client cache hasn't been updated to include them yet. When the client cache is updated we'll get an event that triggers this reconciler, so no requeue or error is needed. I'll update the comment to explain

api/v1alpha1/nnf_node_storage_types.go
matthew-richerson: We were converting from lowercase in the directivebreakdown/servers resource where they're used as labels to uppercase for the NnfStorage/NnfNodeStorage. We were doing that because nnf-ec wanted them as upper case for its arguments. We don't need them as uppercase anymore with nnf-ec out of the picture for Lustre, so I changed it to keep the case consistent in all our resources.

Flux shouldn't be looking at the NnfStorage or NnfNodeStorage, and it shouldn't be aware of Lustre targets at all. The directivebreakdown and servers resources haven't changed in this area.

I'll update the samples.

api/v1alpha1/nnfstorageprofile_types.go
matthew-richerson: I should add it. It wasn't plumbed in to Clientmountd before, but it should be pretty straightforward to get it working now.

roehrich-hpe:
I can see where `CommandArgs.Mkfs` is set, but I have not yet found a place where `CommandArgs.{Mount,MountTarget}` are set.

api/v1alpha1/nnfstorageprofile_types.go
roehrich-hpe: Right, we never did use MountCompute (oops).  But we were using MountRabbit, and that's been lost.

roehrich-hpe:
I'm thinking I ought to be able to deploy a KIND env, and run the clientmountd on my mac, pointed at the k8s env.   Have you tried that?

pkg/command/cmd.go
```diff
@@ -59,16 +58,16 @@ func RunWithTimeout(args string, timeout int) (string, error) {
 	return stdout.String(), nil
 }
 
-func Run(args string) (string, error) {
+func Run(args string, log logr.Logger) (string, error) {
 	timeoutString, found := os.LookupEnv("NNF_COMMAND_TIMEOUT_SECONDS")
```
bdevcich, line 62: Was this never set before?

api/v1alpha1/nnfstorageprofile_types.go
```diff
@@ -35,14 +35,19 @@ type NnfStorageProfileLustreCmdLines struct {
 	// Use the --mkfsoptions argument to specify the zfs create options.  See zfsprops(7).
 	// Use the --mountfsoptions argument to specify persistent mount options for the lustre targets.
 	Mkfs string `json:"mkfs,omitempty"`
+
+	// MountTarget specifies the mount command line for the lustre target.
+	// For persistent mount options for lustre targets, do not use this array; use the --mountfsoptions
+	// argument to mkfs.lustre instead.
+	MountTarget string `json:"mountTarget,omitempty"`
```
roehrich-hpe, line 42: I'm not clear on the different purposes of the MountTarget here, and the MountTarget in the next struct.

api/v1alpha1/nnfstorageprofile_types.go
```diff
@@ -115,6 +120,12 @@ type NnfStorageProfileLustreData struct {
...
 	OstOptions NnfStorageProfileLustreMiscOptions `json:"ostOptions,omitempty"`
+
+	// MountRabbit specifies mount options for mounting on the Rabbit.
+	MountRabbit string `json:"mountRabbit,omitempty"`
+
+	// MountCompute specifies mount options for mounting on the Compute.
+	MountCompute string `json:"mountCompute,omitempty"`
 }
 
 // NnfStorageProfileCmdLines defines commandlines to use for mkfs, and other utilities.
```
roehrich-hpe, line 131: Update this comment to clarify that these are on the non-lustre commandlines.  I had to go back and forth between these two structs a few times to identify the difference between the two structs that define MountRabbit/MountCompute.

matthew-richerson: whoops I pushed this to the wrong branch. I'm still working on these changes.

pkg/command/cmd.go
matthew-richerson: The timeout? This Run() function is new with my code. It wasn't set until now. There's a separate timeout env variable for commands run from nnf-ec

matthew-richerson:
> I'm thinking I ought to be able to deploy a KIND env, and run the clientmountd on my mac, pointed at the k8s env. Have you tried that?

I haven't tried it, but you could with a little bit of work. That Clientmount reconciler in DWS we were talking about in stand up this morning runs in kind to fake out client mounts when Clientmountd isn't running. You'd have to disable that.

api/v1alpha1/nnfstorageprofile_types.go
matthew-richerson: okay it's been tested now. I fixed the two things you commented on so far

internal/controller/nnf_clientmount_controller.go
roehrich-hpe: This fake NnfNodeStorage is not getting the `nnf.cray.hpe.com/pinned_storage_profile_name*` labels.  They aren't on the donor ClientMount that it's using.  So it can't look up the pinned storage profile, to find the mount options it should use.

My workarea is a few commits behind the PR.  I'm using git rev c8260fe.

matthew-richerson: This is what I see for labels on a compute node Clientmount:
```
    dataworkflowservices.github.io/owner.kind: NnfAccess
    dataworkflowservices.github.io/owner.name: global-0-computes
    dataworkflowservices.github.io/owner.namespace: default
    dataworkflowservices.github.io/workflow.name: global
    dataworkflowservices.github.io/workflow.namespace: default
    dataworkflowservices.github.io/workflow.uid: 79783957-f8f1-479f-a056-b160661a95aa
    nnf.cray.hpe.com/directive_index: "0"
    nnf.cray.hpe.com/pinned_storage_profile_name: global-0
    nnf.cray.hpe.com/pinned_storage_profile_namespace: default
    nnf.cray.hpe.com/target_directive_index: "0"
```

That hasn't changed for a while, so I'm not sure why you're seeing that. What's your #DW line?

roehrich-hpe: My #DW lines look like this:
```
    - "#DW jobdw type=xfs capacity=1GB name=stg1"
    - "#DW copy_out source=$DW_JOB_stg1 destination=/lus/foo"
```
But I'm also running the clientmountd on my mac...so I guess I still don't have that working correctly.

pkg/blockdevice/zpool.go
```diff
@@ -113,15 +113,20 @@ func (z *Zpool) GetDevice() string {
...
+		if strings.Contains(err.Error(), "dataset does not exist") {
+			return false, nil
+		}
+
+		return false, fmt.Errorf("could not run zfs to check for zpool device %w", err)
 	}
 
 	if len(output) == 0 {
-		return false
+		return false, fmt.Errorf("zfs get returned no output")
```
bdevcich, line 128: You might consider quoting the `zfs get`, as it reads a bit funny without it: 

```suggestion
		return false, fmt.Errorf("'zfs get' returned no output")
```

roehrich-hpe (approved):

bdevcich (approved):

PR #254, opened 2024-01-31
Let the SystemConfiguration controller add taints and labels to rabbits

roehrich-hpe:
This controller is now responsible for adding the taints and labels to the
rabbit nodes.  It does this in three passes.

Pass 1 adds cray.nnf.node label and NoSchedule taint to each rabbit.  It skips
any rabbit that is already marked with the cray.nnf.node.cleared label, which
marks that rabbit as finished.  If any rabbit is modified in this pass then
it asks to re-enter the reconciler before going to the next pass.

Pass 2 adds the NoExecute taint to each rabbit.  Again, it skips any rabbit
that is already maked with the cray.nnf.node.cleared label.  This taint will
evict any pod that isn't supposed to be on a rabbit; unfortunately, it will
also evict everything else, but we'll open it for scheduling again in the
next pass.

Pass 3 adds the cray.nnf.node.cleared label and removes the NoExecute taint.
This is the final step for each rabbit.

In Pass 2 we ensure that all rabbits get the NoExecute taint together, except
those rabbits that are marked as already finished (with the
cray.nnf.node.cleared label).  This means that the pods we're trying to evict
will be squeezed out of the rabbits; by this time all rabbits will
have the NoSchedule taint from Pass 1, so we won't be playing whack-a-mole with
those pods.


roehrich-hpe:
The nnf-deploy PR is https://github.com/NearNodeFlash/nnf-deploy/pull/121

internal/controller/nnf_systemconfiguration_controller.go
```diff
@@ -190,12 +205,130 @@ func (r *NnfSystemConfigurationReconciler) createNamespaces(ctx context.Context,
...
+			}
+			labels := node.GetLabels()
+			if labels == nil {
+				labels = make(map[string]string)
+			}
+
+			if pass > 0 {
+				// On this node, do later passes only if we haven't already
+				// completed all passes here.
+				if _, present := labels["cray.nnf.node.cleared"]; present {
```
ajfloeder, line 264: Looks like this label sticks around for the case when new nodes are added. I'm not sure I'll remember what `cleared` means. Maybe `taintsAndLabelsComplete`?

bdevcich, line 264: The label is used in multiple places - it should be a constant.

internal/controller/nnf_systemconfiguration_controller.go
```diff
@@ -190,12 +205,130 @@ func (r *NnfSystemConfigurationReconciler) createNamespaces(ctx context.Context,
...
+					continue
+				}
+			}
+
+			taint := &corev1.Taint{
+				Key:    "cray.nnf.node",
```
bdevcich, line 270: `cray.nnf.node` should be a const, too

internal/controller/nnf_systemconfiguration_controller.go
```diff
@@ -190,12 +205,130 @@ func (r *NnfSystemConfigurationReconciler) createNamespaces(ctx context.Context,
...
+				Value:  "true",
+				Effect: taintEffectPerPass[pass],
```
bdevcich, line 272: Couldn't you do `Effect: effect` here instead?

internal/controller/nnf_systemconfiguration_controller.go
```diff
@@ -190,12 +205,130 @@ func (r *NnfSystemConfigurationReconciler) createNamespaces(ctx context.Context,
...
+				// SystemConfiguration hasn't been updated.
+				log.Info("unable to find node", "error", err)
+				continue
+			}
+			labels := node.GetLabels()
+			if labels == nil {
+				labels = make(map[string]string)
+			}
+
+			if pass > 0 {
```
bdevcich, line 261: This block is a little confusing to me. Couldn't you say `if effect != corev1.TaintEffectNoSchedule` and then make the comment that if the cleared label is already present then there's nothing more to be done? Do I have that right?

bdevcich (approved):

internal/controller/nnf_systemconfiguration_controller.go
roehrich-hpe: I've put in some constants.  We've been abusing this particular string, using it for at least 3 different purposes.  I spent some time working to have a unique string for each of those purposes but that rabbit hole goes pretty deep into this repo and other repos so I decided to back off and come up for air.

ajfloeder (approved):

PR #255, opened 2024-02-01
Remove the ComputeNodes array from SystemConfiguration

roehrich-hpe:
The list of compute node names is already found within the spec.StorageNodes array. By having the names in just one array, rather than two, that ought to help reduce errors.

However, the main problem was this: The spec.ComputeNodes array can account for up to 75% of the size of the resource on a large cluster.  This has caused problems for 'kubectl apply' (client-side-apply) when it creates the "last-applied-configuration" annotation for the resource, with the serialized version of the resource exceeding the maximum allowed size of an annotation.

Add spec.ExternalComputeNodes to contain a list of compute nodes that are not directly matched with any of the nodes in the StorageNodes list.


ajfloeder (approved):

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/NearNodeFlash/nnf-sos
 
 go 1.19
 
+replace github.com/DataWorkflowServices/dws => ../dws
+
```
bdevcich, line 6: Reminder

bdevcich (approved):

PR #257, opened 2024-02-07
Add namespace to index mount directories to avoid DM collision

bdevcich:
Use unique index mount directories that consist of
`<namespace>-<index>`. These unique directories guard against potential
data loss when doing data movement operations. Having the namespace
(i.e. rabbit name) included in the mount path ensures that these
individual compute mount points (index mount directories) are always
unique.

For example, on a 2 rabbit job with 1 compute node each, you would see
the following directories across the 2 rabbits:

     /mnt/nnf/7b5dda61-9d91-4b50-a0d3-f863d0aac25b-0/rabbit-node-1-0
     /mnt/nnf/7b5dda61-9d91-4b50-a0d3-f863d0aac25b-0/rabbit-node-2-0

If you did not include the namespace, then the paths would be:

     /mnt/nnf/7b5dda61-9d91-4b50-a0d3-f863d0aac25b-0/0
     /mnt/nnf/7b5dda61-9d91-4b50-a0d3-f863d0aac25b-0/0

On a copy_out directive, if the user specifies the entire directory as
the source (e.g. `source=$DW_JOB_MY_GFS2`), then data movement will pick
up each of those `0/` directories and drop them in the same spot in the
destination. This would lead to potential data loss.

### DIFF
--- a/internal/controller/nnf_access_controller.go
+++ b/internal/controller/nnf_access_controller.go
@@ -1021,7 +1021,7 @@ func clientMountName(access *nnfv1alpha1.NnfAccess) string {
 // For rabbit mounts, use unique index mount directories that consist of <namespace>-<index>.  These
 // unique directories guard against potential data loss when doing copy out data movement
 // operations. Having the namespace (rabbit name) included in the mount path ensures that these
-// individual compute mount points are always unique.
+// individual compute mount points are unique.
 //
 // Ex:
 //
@@ -1034,7 +1034,7 @@ func clientMountName(access *nnfv1alpha1.NnfAccess) string {
 //	/mnt/nnf/7b5dda61-9d91-4b50-a0d3-f863d0aac25b-0/0
 //
 // When data movement copies the MountPathPrefix to global lustre, then the contents of these
-// directories are merged together and potential data loss can occur if the contents do not have
+// directories are merged together and data loss can occur if the contents do not have
 // unique filenames.
 func getIndexMountDir(namespace string, index int) string {
 	return fmt.Sprintf("%s-%s", namespace, strconv.Itoa(index))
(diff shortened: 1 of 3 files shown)


internal/controller/nnf_access_controller.go
```diff
@@ -1017,6 +1018,28 @@ func clientMountName(access *nnfv1alpha1.NnfAccess) string {
 	return access.Namespace + "-" + access.Name
 }
 
+// For rabbit mounts, use unique index mount directories that consist of <namespace>-<index>.  These
+// unique directories guard against potential data loss when doing copy out data movement
+// operations. Having the namespace (rabbit name) included in the mount path ensures that these
+// individual compute mount points are always unique.
```
ajfloeder, line 1024:
```suggestion
// individual compute mount points are unique.
```

internal/controller/nnf_access_controller.go
```diff
@@ -1017,6 +1018,28 @@ func clientMountName(access *nnfv1alpha1.NnfAccess) string {
...
+//	/mnt/nnf/7b5dda61-9d91-4b50-a0d3-f863d0aac25b-0/rabbit-node-1-0
+//	/mnt/nnf/7b5dda61-9d91-4b50-a0d3-f863d0aac25b-0/rabbit-node-2-0
+//
+// If you did not include the namespace, then the paths would be:
+//
+//	/mnt/nnf/7b5dda61-9d91-4b50-a0d3-f863d0aac25b-0/0
+//	/mnt/nnf/7b5dda61-9d91-4b50-a0d3-f863d0aac25b-0/0
+//
+// When data movement copies the MountPathPrefix to global lustre, then the contents of these
+// directories are merged together and potential data loss can occur if the contents do not have
```
ajfloeder, line 1037:
```suggestion
// directories are merged together and data loss can occur if the contents do not have
```

ajfloeder (approved):

roehrich-hpe (approved):

roehrich-hpe:
No, don't force-push.  Then we can't diff before-and-after.

bdevcich:
> No, don't force-push. Then we can't diff before-and-after.

I rebased from master. No changes.

PR #259, opened 2024-02-09
zpool destroy: return success if the pool doesn't exist

bdevcich:
Tested by:
1. Update the `zpoolCreate` options for mdt in the nnfstorageprofile to include
   bad options
2. Without this change, setup will hang on the create. Transition to
   teardown and it will also hang since it can't remove the pool that
   doesn't exist.
3. With this change, teardown will complete since destroy has nothing to
   remove.


ajfloeder (approved):

bdevcich:
I originally had this coded up with an Exists() function to be used by both Create and Destory, but this is much simplier.

roehrich-hpe (approved):

PR #261, opened 2024-02-13
NnfDataMovementManager: Add UpdateStrategy to support parallel rolling updates

bdevcich:
The NNF DMM manages the nnf-dm-worker DaemonSet, which is reponsible for
the nnf-dm-worker pods. Adding UpdateStrategy to the
NnfDataMovementManagerSpec ensures admins can configure the update
strategy for those pods.

This allows for a non-sequential update when rolling out new worker
pods.


api/v1alpha1/nnf_datamovementmanager_types.go
```diff
@@ -50,6 +51,10 @@ type NnfDataMovementManagerSpec struct {
 	// manages the per node data movement operations.
 	Template corev1.PodTemplateSpec `json:"template"`
 
+	// UpdateStrategy defines the UpdateStrategy that is used for the basis of the worker Daemon Set
+	// that manages the per node data movement operations.
+	UpdateStrategy appsv1.DaemonSetUpdateStrategy `json:"updateStrategy"`
```
roehrich-hpe, line 56: Does it need to be required?  Can it be omitempty?

ajfloeder (approved):

roehrich-hpe (approved):

api/v1alpha1/nnf_datamovementmanager_types.go
roehrich-hpe: Never mind, I see that sos has it covered always.

bdevcich: I originally had it as optional but then changed my mind to just make it a requirement. We basically never want it to use the default of 1 anyway, so I think expecting this field is the way to go.

PR #266, opened 2024-02-19
Xfs failure

ajfloeder:
From a high level the issue here is that the nnf_node_block_storage resource came to the ready state prematurely allowing the nnf_node_storage to take off and attempt to create the pv's/vg's/lv's. 

```bash
make simple
...
Ran 4 of 26 Specs in 1679.888 seconds
SUCCESS! -- 4 Passed | 0 Failed | 1 Pending | 21 Skipped
PASS

Ginkgo ran 1 suite in 28m6.684407332s
Test Suite Passed
```


internal/controller/nnf_node_block_storage_controller.go
```diff
@@ -342,16 +342,14 @@ func (r *NnfNodeBlockStorageReconciler) createBlockDevice(ctx context.Context, n
 			if _, ok := allocationStatus.Accesses[nodeName]; !ok {
 				log.Info("Created storage group", "Id", storageGroupId)
 				allocationStatus.Accesses[nodeName] = nnfv1alpha1.NnfNodeBlockStorageAccessStatus{StorageGroupId: sg.Id}
-
```
ajfloeder, line 345: If we exit the loop here, the nnf_node_block_storage resource prematurely becomes ready.

go.mod
```diff
@@ -22,7 +22,7 @@ require (
 	k8s.io/api v0.28.1
 	k8s.io/apimachinery v0.28.1
 	k8s.io/client-go v0.28.1
-	k8s.io/kubernetes v1.28.4
+	k8s.io/kubernetes v1.15.0-alpha.0
```
roehrich-hpe, line 25: That looks wrong.
Be careful about using the `-u` option on `go get`.  I generally don't add that option unless I explicitly want to do a general dependency update.

roehrich-hpe (approved):

roehrich-hpe (approved):

PR #271, opened 2024-02-26
Retry failing wipefs to give mapper device time to appear

ajfloeder:
nnf-node-storage controller calls `wipefs` to determine whether a filesystem has been created on a logical volume (lv). The logical volume must first be activated for it to appear in the `/dev/mapper` directory which is where `wipefs` looks for its device. The time it takes for the activate to start and the device to appear in the `/dev/mapper` directory can be a bit longer than the time it takes for the reconciler to `activate` then `wipefs` the logical volume.

This change adds a retry loop around the `wipefs` command to retry at 1 second intervals for a total of 10 seconds before failing the `wipefs` command.

### DIFF
--- a/pkg/blockdevice/lvm.go
+++ b/pkg/blockdevice/lvm.go
@@ -22,6 +22,8 @@ package blockdevice
 import (
 	"context"
 	"fmt"
+	"os"
+	"strconv"
 	"strings"
 	"time"
 
@@ -237,10 +239,23 @@ func (l *Lvm) GetDevice() string {
 // CheckFormatted determines if the device has been formatted
 func (l *Lvm) CheckFormatted() (bool, error) {
 
-	// HACK: Give the device 10 seconds to appear in /dev/mapper
+	// Default to 10 second timeout
+	retryPeriod := 10 * time.Second
+
+	// Look for environment variable to override
+	timeoutString, found := os.LookupEnv("NNF_WIPEFS_MAPPER_WAIT_COMMAND_TIMEOUT")
+	if found {
+		timeout, err := strconv.Atoi(timeoutString)
+		if err == nil {
+			if timeout > 0 {
+				retryPeriod = time.Duration(timeout) * time.Second
+			}
+		}
+	}
+
+	// Give the device some time to appear in /dev/mapper
 	var err error
 	var output string
-	retryPeriod := 10 * time.Second
 
 	// The /dev/mapper device can take a while.
 	// Retry failing wipefs to give it time.
--- a/pkg/blockdevice/lvm.go
+++ b/pkg/blockdevice/lvm.go
@@ -254,17 +254,24 @@ func (l *Lvm) CheckFormatted() (bool, error) {
 	}
 
 	// Give the device some time to appear in /dev/mapper
+	// Retry failing wipefs while we wait
 	var err error
 	var output string
-
-	// The /dev/mapper device can take a while.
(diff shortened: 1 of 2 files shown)


pkg/blockdevice/lvm.go
```diff
@@ -210,19 +214,52 @@ func (l *Lvm) Deactivate(ctx context.Context) (bool, error) {
...
+// CheckFormatted determines if the device has been formatted
 func (l *Lvm) CheckFormatted() (bool, error) {
-	output, err := command.Run(fmt.Sprintf("wipefs --noheadings --output type %s", l.GetDevice()), l.Log)
-	if err != nil {
-		return false, fmt.Errorf("could not run wipefs to determine if device is formatted: %w", err)
-	}
 
-	if len(output) == 0 {
-		return false, nil
+	// HACK: Give the device 10 seconds to appear in /dev/mapper
```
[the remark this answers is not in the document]
bdevcich: Or perhaps in the storage profile?

bdevcich (approved):

bdevcich (approved):

pkg/blockdevice/lvm.go
```diff
@@ -210,19 +216,65 @@ func (l *Lvm) Deactivate(ctx context.Context) (bool, error) {
...
+	// Give the device some time to appear in /dev/mapper
+	var err error
+	var output string
+
+	// The /dev/mapper device can take a while.
+	// Retry failing wipefs to give it time.
+	device := l.GetDevice()
+	for start := time.Now(); time.Since(start) < retryPeriod; time.Sleep(time.Second) {
+
+		output, err = command.Run(fmt.Sprintf("wipefs --noheadings --output type %s", device), l.Log)
```
behlendorf, line 265: I'd consider adding a `stat()` call before `command.Run()` to verify the path exists and we should expect `wipefs` to succeed.

ajfloeder: I considered that and in fact coded it. It seems we need to poll on some command, be it `stat` or `wipefs`. Is there a benefit to waiting with `stat()` above waiting with `wipefs`? In either case, if the path we are waiting for fails to show up, we are left with an error.

roehrich-hpe: Tony, you can use `strings.Contains(err, "mumble")` to verify that wipefs failed because the device isn't present.  Then "mumble" would be just a snippet from the error, such as "not found".  Any other error and maybe we shouldn't be looping.

As it's coded now, if the first error in the loop and the last error in the loop are different, and both have nothing to do with "not found", then we'll never know the first error.  We should probably know that first error.

roehrich-hpe (approved):

pkg/blockdevice/lvm.go
behlendorf: The advantage of checking if the files exists first is you can do it with a system call.  In the error case this entirely avoids the fork/exec and any execution `wipefs` might need to do before it fails.  It looks like both the [lstat](https://pkg.go.dev/syscall#Lstat) and [access](https://pkg.go.dev/syscall#Access) syscalls are available.  Since you just need to check for existence, `access(2)` is probably the easiest way to go.

If you're feeling really ambitious `inotify` is the way to handle this on Linux without polling.  If I understand correctly [fsnotify](https://pkg.go.dev/github.com/fsnotify/fsnotify) is the preferred portable wrapper for this in go, but I'm not sure offhand what additional dependencies you might need for that.

PR #280, opened 2024-03-08
Fix trailing slash issues with Data Movement

bdevcich:
- Moved `getRabbitRelativePath` to its own function for testability
- `getRabbitRelativePath` now preserves trailing slashes for the path
- Fixed issue with `splitStagingArgumentIntoNameAndPath` assuming a
  leading slash
- Added unit tests for `splitStagingArgumentIntoNameAndPath` and `getRabbitRelativePath`

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -21,12 +21,12 @@
             "type": "go",
             "request": "launch",
             "mode": "test",
-            "program": "${relativeFileDirname}",
+            "program": "${workspaceFolder}/internal/controller",
             "args": [
                 "-ginkgo.v",
             ],
             "env": {
-                "KUBEBUILDER_ASSETS": "${workspaceFolder}/bin/k8s/1.26.0-darwin-amd64",
+                "KUBEBUILDER_ASSETS": "${workspaceFolder}/bin/k8s/1.28.0-darwin-arm64",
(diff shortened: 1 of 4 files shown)


bdevcich:
This is ready for action now. It passes that tests that it should - no trailing slash on source. It does leave open an issue with a trailing slash on the source since this fix **makes sure that trailing slashes are preserved**.

So, if you do the following:

```
copy_out src=\$DW_JOB_my_gfs2 dest=/lus/global/user
copy_out src=\$DW_JOB_my_gfs2 dest=/lus/global/user/
```

Then you will get index mount directories at `/lus/global/user`:

```
/lus/global/user/rabbit-node-1-0/file.out
...
```

Now that trailing slashes are preserved, we don't get that if you add a trailing slash on the source. This is because `dcp` is now copying the contents of the index mount directory rather than the directory itself. So with trailing slashes on the source:

```
copy_out src=\$DW_JOB_my_gfs2/ dest=/lus/global/user
copy_out src=\$DW_JOB_my_gfs2/ dest=/lus/global/user/
```

You would just get:

```
/lus/global/user/data.out
```

To address this, another PR will be opened to create the index mount directories in the destination to ensure that every copy out source has a safe spot to land and avoid collision/data loss.

ajfloeder (approved):

PR #283, opened 2024-03-13
Clientmount issues

ajfloeder:
In the SimpleFileSystem code when an unmount operation is successful, don't attempt unmount activities on subsequent Reconciles since the filesystem has already been unmounted. That is, trust the `Ready` status of the clientmount resource.

### DIFF
--- a/internal/controller/nnf_clientmount_controller.go
+++ b/internal/controller/nnf_clientmount_controller.go
@@ -234,7 +234,7 @@ func (r *NnfClientMountReconciler) changeMount(ctx context.Context, clientMount 
 			}
 		}
 	} else {
-		unmounted, err := fileSystem.Unmount(ctx, clientMountInfo.MountPath)
+		unmounted, err := fileSystem.Unmount(ctx, clientMountInfo.MountPath, clientMount.Status.Mounts[index].Ready)
 		if err != nil {
 			return dwsv1alpha2.NewResourceError("unable to unmount file system").WithError(err).WithMajor()
 		}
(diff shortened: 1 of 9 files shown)


pkg/filesystem/filesystem.go
```diff
@@ -38,7 +38,7 @@ type FileSystem interface {
 	Mount(ctx context.Context, path string, complete bool) (bool, error)
 
 	// Unmount the file system
-	Unmount(ctx context.Context, path string) (bool, error)
+	Unmount(ctx context.Context, path string, complete bool) (bool, error)
```
ajfloeder, line 41: Add the parameter that indicates whether we've already successfully unmounted to the API.

internal/controller/nnf_clientmount_controller.go
```diff
@@ -234,7 +234,7 @@ func (r *NnfClientMountReconciler) changeMount(ctx context.Context, clientMount
 			}
 		}
 	} else {
-		unmounted, err := fileSystem.Unmount(ctx, clientMountInfo.MountPath)
+		unmounted, err := fileSystem.Unmount(ctx, clientMountInfo.MountPath, clientMount.Status.Mounts[index].Ready)
```
roehrich-hpe, line 237: Two questions:
1. Why bother to call this fileSystem.Unmount() if Ready==True?
2. Reconcilers are built around the idea that they "reconcile"--they always verify and correct their Status. If you're going to deviate from that model, then that ought to be described and justified.

roehrich-hpe (changes requested):

internal/controller/nnf_clientmount_controller.go
roehrich-hpe: 3. If the status update that was supposed to set Ready=True hits a stale cache error (I forget the error about the cache being stale), then you'll be back here repeating a real unmount anyway. Will that stuff things up again?

PR #285, opened 2024-03-22
Upgrade controller-gen to v0.14.0 and setup-envtest to 3/20/24

roehrich-hpe:
The k8s-go community broke a few things this morning, our time.


roehrich-hpe:
The switch to controller-gen v0.14.0 is something we can keep.  The pinning of setup-envtest feels like something that ought to be short-term; like maybe the k8s-go folks will revert that breakage by Monday.

roehrich-hpe:
The setup-envtest I picked is from two days ago, since the latest one today is broken. Here's where I'm looking for setup-envtest versions:
https://pkg.go.dev/sigs.k8s.io/controller-runtime/tools/setup-envtest/versions?tab=versions

bdevcich (approved):

PR #286, opened 2024-03-25
Ensure NnfNodeBlockStorages for lustre have a capacity of at least 4GB

bdevcich:

### DIFF
--- a/internal/controller/nnf_storage_controller.go
+++ b/internal/controller/nnf_storage_controller.go
@@ -354,7 +354,8 @@ func (r *NnfStorageReconciler) createNodeBlockStorage(ctx context.Context, nnfSt
 
 				for i := range nnfNodeBlockStorage.Spec.Allocations {
 
-					// For lustre - bump up the capacity if less than the floor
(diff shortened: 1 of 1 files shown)


internal/controller/nnf_storage_controller.go
```diff
@@ -347,7 +353,13 @@ func (r *NnfStorageReconciler) createNodeBlockStorage(ctx context.Context, nnfSt
 				}
 
 				for i := range nnfNodeBlockStorage.Spec.Allocations {
-					nnfNodeBlockStorage.Spec.Allocations[i].Capacity = allocationSet.Capacity
+
+					// For lustre - bump up the capacity if less than the floor
```
ajfloeder, line 357: Maybe mention zpool here.

ajfloeder (approved):

internal/controller/nnf_storage_controller.go
bdevcich: How's this?

```
// For lustre (zfs), bump up the capacity if less than the floor. This is to
// ensure that zpool create does not fail if the size is too small.
```

PR #287, opened 2024-03-25
Add `zpool import` on Node EC Data controller startup

bdevcich:
This ensures that any existing pools are imported and present after a reboot of the rabbit node.

Tested by:
1. Create lustre workflow and transition to PostRun
2. Reboot the rabbit-node prior to Teardown
3. Once rebooted, verify that zool list includes pools
4. Verify log contains "Imported all available zpools"

1. Do #1 from above.
2. Transition to Teardown and then very quickly reboot the rabbit node (`ssh rabbit-node-2 reboot`).
3. Once rabbit node reboots, workflow will transition to Teardown complete.
4. Verify #4 above.
5. Verify `zpool list` as no pools left.

moved import to NnfNodeECData.Start()


bdevcich:
Didn't mean to open this yet - I'd like to retest first.

PR #296, opened 2024-04-22
SystemConfiguration controller should not delete namespaces

roehrich-hpe:
On a large system we do not want the time & traffic hit that comes with deleting namespaces.

### DIFF
--- a/internal/controller/nnf_systemconfiguration_controller.go
+++ b/internal/controller/nnf_systemconfiguration_controller.go
@@ -124,12 +124,6 @@ func (r *NnfSystemConfigurationReconciler) Reconcile(ctx context.Context, req ct
 		validNamespaces[*name] = struct{}{}
 	}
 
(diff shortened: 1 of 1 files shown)


roehrich-hpe:
> Does this mean compute node namespaces are going to stick around then?

Yes. If you have the full-size SystemConfiguration, and you decide to remove a rabbit and a chunk of computes, then this will  leave the namespaces.

Without this, you would have to wait for the systemconfiguration controller to figure out which namespaces have been deleted and to walk through them doing the deletes--that takes a long time when you're using the full-size SystemConfiguration.

bdevcich (approved):

ajfloeder (commented):
How would Flux know that rabbits/compute nodes have been deleted?

roehrich-hpe:
> How would Flux know that rabbits/compute nodes have been deleted?

It's not doing that by watching k8s namespaces.

ajfloeder (approved):

PR #305, opened 2024-05-07
filesystem-helpers: guard against empty status allocations

bdevcich:
Ran into this when canceling a bunch of flux jobs repeatedly. Something got wedged in regards to persistent storage and a panic occurs. This guards against the two index out of range panics.

### DIFF
--- a/internal/controller/filesystem_helpers.go
+++ b/internal/controller/filesystem_helpers.go
@@ -196,7 +196,9 @@ func newLvmBlockDevice(ctx context.Context, c client.Client, nnfNodeStorage *nnf
 			return nil, dwsv1alpha2.NewResourceError("could not get NnfNodeBlockStorage: %v", client.ObjectKeyFromObject(nnfNodeBlockStorage)).WithError(err).WithUserMessage("could not find storage allocation").WithMajor()
 		}
 
-		devices = nnfNodeBlockStorage.Status.Allocations[index].Accesses[os.Getenv("NNF_NODE_NAME")].DevicePaths
+		if len(nnfNodeBlockStorage.Status.Allocations) > 0 && len(nnfNodeBlockStorage.Status.Allocations[index].Accesses) > 0 {
+			devices = nnfNodeBlockStorage.Status.Allocations[index].Accesses[os.Getenv("NNF_NODE_NAME")].DevicePaths
+		}
 	}
 
 	for _, device := range devices {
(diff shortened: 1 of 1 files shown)


internal/controller/filesystem_helpers.go
```diff
@@ -196,7 +196,9 @@ func newLvmBlockDevice(ctx context.Context, c client.Client, nnfNodeStorage *nnf
 			return nil, dwsv1alpha2.NewResourceError("could not get NnfNodeBlockStorage: %v", client.ObjectKeyFromObject(nnfNodeBlockStorage)).WithError(err).WithUserMessage("could not find storage allocation").WithMajor()
 		}
 
-		devices = nnfNodeBlockStorage.Status.Allocations[index].Accesses[os.Getenv("NNF_NODE_NAME")].DevicePaths
+		if len(nnfNodeBlockStorage.Status.Allocations) > 0 && len(nnfNodeBlockStorage.Status.Allocations[index].Accesses) > 0 {
+			devices = nnfNodeBlockStorage.Status.Allocations[index].Accesses[os.Getenv("NNF_NODE_NAME")].DevicePaths
+		}
```
bdevcich, line 201: Do we need to take a different action if there are no devices? I'm not sure how to proceed in that case. I'm not exactly sure how I even got here, but this fix did allow the workflow to complete teardown instead of a CrashLoopBackoff.

ajfloeder: Was there a failure to allocate namespaces prior to this?

ajfloeder: Without the devices here, it seems like you just exit.

roehrich-hpe: So it couldn't do the teardown because it didn't have allocated devices. But in that case did all of the work in `deleteAllocation()` actually happen, maybe earlier?
Do we know whether it ever had allocated devices?

bdevcich: I'm not sure in either case, I just know the pod hit this case after some behavior that probably shouldn't have happened. I'm not sure which tests ran and which didn't. Sorry I don't have more detail.

bdevcich:
Here's the log from the CLBO pod:
[empty-status-allocations-.log](https://github.com/NearNodeFlash/nnf-sos/files/15249429/empty-status-allocations-.log)

ajfloeder:
> Here's the log from the CLBO pod: [empty-status-allocations-.log](https://github.com/NearNodeFlash/nnf-sos/files/15249429/empty-status-allocations-.log)

That log indicates that just before the pod crashed a storage pool was deleted. I think the code you've added makes sense, but how we got to that point is not clear. The log looks truncated.
Are there other logs on that Rabbit? `/var/log/pods/nnf-system_nnf-node-manager-<uid>/manager`

internal/controller/filesystem_helpers.go
roehrich-hpe: Let's add a log.Error() here and see how things shake out in the future.

roehrich-hpe (approved):

ajfloeder (approved):

PR #307, opened 2024-05-30
Configure time for UTC.

roehrich-hpe:
Configure the time the same way we do for other services. This allows both the NLC and SLC controllers to agree on using UTC, rather than having one local and the other UTC.


roehrich-hpe:
This does not address Brian's request to switch to local time, but it does finally get all of our controllers, from all of our repos, agreeing to use UTC.
We can figure out the local time thing in another PR.

ajfloeder (approved):

matthew-richerson (approved):

bdevcich (approved):

PR #308, opened 2024-05-30
Serialize the NLC reconcilers, take 2

roehrich-hpe:
Add channel semaphores to serialize the startup of the NLC reconcilers that are using EC.  Add locks to each reconciler to block its Reconcile() until it has completed its Start().

This is a rewrite of an earlier version, this time using close() on the channel rather than leaving the channels open.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -182,8 +182,7 @@ func (c *nodeLocalController) SetupReconcilers(mgr manager.Manager, opts *nnf.Op
 
 	// Coordinate the startup of the NLC controllers that use EC.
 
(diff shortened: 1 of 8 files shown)


roehrich-hpe:
I was reading some code in a controller-runtime PR, and I thought this looked like a better way to use channels:
https://github.com/kubernetes-sigs/controller-runtime/pull/2407/files#diff-bfb58f4e52ee206ea6c273b044777426855626395552c33a044eb5a7862e4bb2R228-R246

bdevcich (approved):

PR #312, opened 2024-06-07
Short-lived clientmountd

roehrich-hpe:
Create a non-caching, non-watching, Reconcile-like loop in mount-daemon/main.go that terminates after it has completed all of the ClientMount resources it can find. The code in the original clientmount_controller.go has been moved to a helper library where it can be used by either type of Reconciler.

### DIFF
--- a/mount-daemon/main.go
+++ b/mount-daemon/main.go
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021-2023 Hewlett Packard Enterprise Development LP
+ * Copyright 2021-2024 Hewlett Packard Enterprise Development LP
  * Other additional copyright holders may be indicated within.
  *
  * The entirety of this work is licensed under the Apache License,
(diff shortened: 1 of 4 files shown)


mount-daemon/main.go
```diff
@@ -269,26 +292,45 @@ func main() {
...
+		setupLog.Error(err, "failed client.New")
 		os.Exit(1)
 	}
 
-	fmt.Println(status)
+	var result ctrl.Result
+	lastChance := false
+	namespace := opts.name // The node's namespace is the same as the node's name.
+	for {
+		result, err = reconcileAll(clnt, namespace)
```
bdevcich, line 317: Do we need to check err for any conflict type errors and requeue?

mount-daemon/main.go
```diff
@@ -176,90 +160,129 @@ func createManager(opts *options) (*managerConfig, error) {
...
-		LeaderElection: false,
-		Cache:          cache.Options{DefaultNamespaces: namespaceCache},
-	})
-	if err != nil {
-		setupLog.Error(err, "unable to start manager")
-		os.Exit(1)
+func reconcileAll(clnt client.Client, namespace string) (ctrl.Result, error) {
+	clientMountList := &dwsv1alpha2.ClientMountList{}
+	listOptions := []client.ListOption{
+		client.InNamespace(namespace),
```
bdevcich, line 262: I know flux shouldn't be running multiple workflows on the same compute, but should this be limited to a workflow along with the namespace?

What would happen if there's another clientmount present from an old workflow that didn't clean up and that returns an error? Would this logic find that error and bomb out without doing any work on the _actual_ clientmount that was requested?

PR #314, opened 2024-06-10
Add option for a single shared VG per Rabbit per workflow

matthew-richerson:
This commit adds a new option, sharedVg, in the NnfStorageProfile for allocations backed by an LVM logical volume. This option allows allocations on a Rabbit node that are part of the same workflow to share a volume group. This reduces the overhead of creating allocations and greatly speeds up the Setup and Teardown phases of the workflow.

This commit also changes the way locking is done with lvmlockd. Previously, when the LV was not needed on a node, the LV was deactivated and the VG lockspace was stopped. Now, only the LV is deactivated in most situations. The VG lockspace is only stopped when a node is completely done with a VG and it won't be activated again.

### DIFF
--- a/internal/controller/filesystem_helpers.go
+++ b/internal/controller/filesystem_helpers.go
@@ -183,6 +183,11 @@ func newLvmBlockDevice(ctx context.Context, c client.Client, nnfNodeStorage *nnf
 	lvmDesc := blockdevice.Lvm{}
 	devices := []string{}
 
+	blockIndex := index
+	if nnfNodeStorage.Spec.SharedAllocation {
+		blockIndex = 0
+	}
+
 	if nnfNodeStorage.Spec.BlockReference.Kind == reflect.TypeOf(nnfv1alpha1.NnfNodeBlockStorage{}).Name() {
 		nnfNodeBlockStorage := &nnfv1alpha1.NnfNodeBlockStorage{
 			ObjectMeta: metav1.ObjectMeta{
@@ -196,8 +201,8 @@ func newLvmBlockDevice(ctx context.Context, c client.Client, nnfNodeStorage *nnf
 			return nil, dwsv1alpha2.NewResourceError("could not get NnfNodeBlockStorage: %v", client.ObjectKeyFromObject(nnfNodeBlockStorage)).WithError(err).WithUserMessage("could not find storage allocation").WithMajor()
 		}
 
-		if len(nnfNodeBlockStorage.Status.Allocations) > 0 && len(nnfNodeBlockStorage.Status.Allocations[index].Accesses) > 0 {
-			devices = nnfNodeBlockStorage.Status.Allocations[index].Accesses[os.Getenv("NNF_NODE_NAME")].DevicePaths
+		if len(nnfNodeBlockStorage.Status.Allocations) > 0 && len(nnfNodeBlockStorage.Status.Allocations[blockIndex].Accesses) > 0 {
+			devices = nnfNodeBlockStorage.Status.Allocations[blockIndex].Accesses[os.Getenv("NNF_NODE_NAME")].DevicePaths
 		} else {
 			// FIXME: This scenario has been encountered after rapid transitions from Setup to
 			// Teardown. The reason is not yet known, but we need a way to log and capture this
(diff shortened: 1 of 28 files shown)


internal/controller/filesystem_helpers.go
```diff
@@ -227,9 +232,14 @@ func newLvmBlockDevice(ctx context.Context, c client.Client, nnfNodeStorage *nnf
 		return nil, dwsv1alpha2.NewResourceError("could not get logical volume name").WithError(err).WithMajor()
 	}
 
+	percentVG := 100
+	if nnfNodeStorage.Spec.SharedAllocation {
+		percentVG = 100 / nnfNodeStorage.Spec.Count
```
bdevcich, line 237:
```suggestion
		percentVG = percentVG / nnfNodeStorage.Spec.Count
```

internal/controller/filesystem_helpers.go
```diff
@@ -227,9 +232,14 @@ func newLvmBlockDevice(ctx context.Context, c client.Client, nnfNodeStorage *nnf
 		return nil, dwsv1alpha2.NewResourceError("could not get logical volume name").WithError(err).WithMajor()
 	}
 
+	percentVG := 100
```
bdevcich, line 235: Shouldn't this be pulled from the profile?

internal/controller/filesystem_helpers.go
roehrich-hpe: I guess that suggestion doesn't make sense to me.  Those two '100's are not the same thing.

internal/controller/nnf_clientmount_controller.go
```diff
@@ -245,6 +264,20 @@ func (r *NnfClientMountReconciler) changeMount(ctx context.Context, clientMount
...
 			log.Info("Unmounted file system", "Mount path", clientMountInfo.MountPath)
 		}
+
+		// If this is an unmount on a compute node, we can fully deactivate the block device since we won't use it
+		// again. If this is a rabbit node, we do a minimal deactivation. For LVM this means leaving the lockspace up
+		fullDeactivate := false
+		if r.ClientType == ClientCompute {
+			fullDeactivate = true
+		}
+		deactivated, err := blockDevice.Deactivate(ctx, fullDeactivate)
```
roehrich-hpe, line 274: This is the only place I've found where a full deactivate can happen, and only on the compute node. Did I miss any other cases?

matthew-richerson: Yep, that's it. The VG lockspace stays open on the Rabbit until the VG is destroyed.

internal/controller/filesystem_helpers.go
matthew-richerson: This percentVG value is used to fill in the $PERCENT_VG variable in the "lvcreate" command for the "--extents" option. If there's only one LV on a VG, "$PERCENT_VG" will expand to "100%VG" so that the LV gets all the space on the VG. If they wanted something other than that (I'm not sure that they ever would), they could use a hard coded value other than "$PERCENT_VG". If the VG is shared between multiple LVs, then "$PERCENT_VG" will expand to something like "25%VG" if there were 4 LVs on the VG. The actual percent of the VG depends on the number of compute nodes making an allocation on a Rabbit, and it may differ between Rabbits in a single workflow.

internal/controller/filesystem_helpers.go
matthew-richerson: Yeah, the `percentVG := 100` is setting the value of "$PERCENT_VG" to be 100% in the case where the VG is not shared and there's only a single LV. The `percentVG = 100 / nnfNodeStorage.Spec.Count` calculation is for when the VG is shared and we need to divide the VG into equal sized chunks.

internal/controller/nnf_node_storage_controller.go
```diff
@@ -256,50 +270,73 @@ func (r *NnfNodeStorageReconciler) deleteAllocation(ctx context.Context, nnfNode
...
+		defer func() {
+			_, err = blockDevices[deferIndex].Deactivate(ctx, false)
+			if err != nil {
+				allocationStatus.Ready = false
+			}
+		}()
 	}
 
-	allocationStatus.Ready = true
+	for index, fileSystem := range fileSystems {
```
roehrich-hpe, line 306: Won't this loop be running at the same time as the defer funcs from above are running? Touching the same fields in nnfNodeStorage.Status.Allocations[]?

matthew-richerson: The defer funcs don't get run until we return from the CreateAllocations() function. So this loop happens before that.

roehrich-hpe: Oh, the surrounding function, not the surrounding block. Got it.

bdevcich (approved):

roehrich-hpe (approved):

PR #318, opened 2024-06-25
Add RequiredDaemons list to DirectiveBreakdown

roehrich-hpe:
Define a new keyword for the "#DW jobdw ..." directive to allow the user to specify a list of required daemons. The new keyword is "requires=" and it takes a list of strings:

  requires=copy-offload,future-thing

Set the RequiredDaemons field in the DirectiveBreakdownStatus with the list of specified daemons.


internal/controller/directivebreakdown_controller.go
```diff
@@ -532,6 +525,10 @@ func (r *DirectiveBreakdownReconciler) populateStorageBreakdown(ctx context.Cont
 	}
 
 	dbd.Status.Storage.AllocationSets = allocationSets
+
+	if wordList, present := argsMap["requires"]; present {
+		dbd.Status.RequiredDaemons = strings.Split(wordList, ",")
```
matthew-richerson, line 530: RequiredDaemons won't get set for "#DW persistentdw". Do we have any reason to limit copy-offload to jobdw only?

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/NearNodeFlash/nnf-sos
 
 go 1.19
 
+replace github.com/DataWorkflowServices/dws => ../dws
+
 require (
```
bdevcich, line 7: Just a reminder to update this after DWS is merged

internal/controller/directivebreakdown_controller.go
roehrich-hpe: I've forgotten how 'persistentdw` is used. I thought it just asks for the volume to be mounted, and had nothing to do with running a compute job.

bdevcich (approved):
This looks good, but I'm curious about `persistentdw`.

internal/controller/directivebreakdown_controller.go
matthew-richerson: "create_persistent" creates the persistent file system on the Rabbit nodes, and it doesn't require any computes to do that. 

"persistentdw" mounts the existing persistent file system from the Rabbits onto compute nodes so it can be used by a job application. There is a user application run as part of this job just like in a jobdw. I can't think of any reason to prevent users from initiating copy-offload with this type of job.

"destroy_persistent" deletes the persistent file system on the Rabbit nodes and doesn't require any computes.

roehrich-hpe: ok, persistentdw is working now.

matthew-richerson (approved):

bdevcich (approved):

PR #331, opened 2024-07-15
 Add new options to NnfStorageProfile

matthew-richerson:
CapacityScalingFactor: An option to increase the capacity in the directive breakdown from what the user requested on the #dw line. This can be used to make the usable file system space more closely match the user's request. An admin must set this to a value that makes sense with the block device and file system options used. This is separately configurable for each file system type.

StorageLabels: A set of labels to add to the labels constraint in the directive breakdown. This is separately configurable for each file system type and each Lustre target type.

I also changed the #dw create_persistent directive to no longer require the "capacity" argument when it's being used to create standalone MGTs.


api/v1alpha1/nnfstorageprofile_types.go
```diff
@@ -188,18 +196,42 @@ type NnfStorageProfileLVMLvChangeCmdLines struct {
 type NnfStorageProfileGFS2Data struct {
 	// CmdLines contains commands to create volumes and filesystems.
 	CmdLines NnfStorageProfileCmdLines `json:"commandlines,omitempty"`
+
+	// Storagelabels defines a list of labels that are added to the DirectiveBreakdown
```
bdevcich, line 200:
```suggestion
	// StorageLabels defines a list of labels that are added to the DirectiveBreakdown
```

api/v1alpha1/nnfstorageprofile_types.go
```diff
@@ -188,18 +196,42 @@ type NnfStorageProfileLVMLvChangeCmdLines struct {
...
+	// +kubebuilder:default:="1.0"
+	CapacityScalingFactor string `json:"capacityScalingFactor,omitempty"`
 }
 
 // NnfStorageProfileRawData defines the Raw-specific configuration
 type NnfStorageProfileRawData struct {
 	// CmdLines contains commands to create volumes and filesystems.
 	CmdLines NnfStorageProfileCmdLines `json:"commandlines,omitempty"`
+
+	// Storagelabels defines a list of labels that are added to the DirectiveBreakdown
```
bdevcich, line 228: and here

config/dws/nnf-ruleset.yaml
```diff
@@ -46,7 +46,7 @@ spec:
       - key: "^capacity$"
         type: "string"
         pattern: "^\\d+(KiB|KB|MiB|MB|GiB|GB|TiB|TB)$"
-        isRequired: true
+        isRequired: false
```
bdevcich, line 49: How does this affect when you ask for xfs with a capacity of 0 or something similar?

roehrich-hpe (approved):

config/dws/nnf-ruleset.yaml
matthew-richerson: That's handled in the persistent storage reconciler. It makes sure that every directive has capacity specified except when it's lustre and StandaloneMgtPoolName is specified. I guess my error message for the second case is wrong though. It shouldn't mention "lustre".

```
	if argsMap["type"] == "lustre" && len(pinnedProfile.Data.LustreStorage.StandaloneMGTPoolName) > 0 {
		if _, exists := argsMap["capacity"]; exists {
			return ctrl.Result{}, dwsv1alpha2.NewResourceError("").WithUserMessage("creating persistent MGT does not accept 'capacity' argument").WithFatal().WithUser()
		}
		...
	} else {
		if _, exists := argsMap["capacity"]; !exists {
			return ctrl.Result{}, dwsv1alpha2.NewResourceError("").WithUserMessage("creating persistent Lustre requires 'capacity' argument").WithFatal().WithUser()
		}
	}
```

Your original question of 0 capacity made me want to try that. We don't handle that case well for jobdw or create_persistent. It just hangs in proposal state. Maybe I'll add on a fix for that bug to this PR since it's in the area.

bdevcich (approved):

PR #335, opened 2024-07-17
Get missing events from k8s Node resource

matthew-richerson:
The Storage reconciler wasn't running when the k8s node resource changed. This is because the enqueuing function wasn't specifying the namespace in the namespacedname struct.

Also, fix the Requeue() function so that it specifies a requeue in the result. In the delete path, the Requeue=true result doesn't always result in a requeue. Use RequeueAfter=1sec instead. This was only an issue when the workflow was deleted without waiting for teardown to complete first. Teardown normally removes the container and port resources that were causing the issue.


ajfloeder (approved):

roehrich-hpe (approved):

roehrich-hpe:
Update the copyright in nnf_workflow_controller_helpers.

bdevcich (approved):
Nice catch

PR #340, opened 2024-07-31
Add NnfLustreMGT resource for tracking fsname

matthew-richerson:
Create a new NnfLustreMGT resource that tracks used fsnames and erases unused fsnames on internal MGTs. A NnfLustreMGT resource is created automatically whenever a Lustre MGT is created. Lustre file systems that want to use the MGT can add a claim to the NnfLustreMGT resource to request an fsname. The NnfLustreMGT controller will find the next available fsname and assign it. When a Lustre file system is torn down, removing the claim from the NnfLustreMGT resource will trigger the NnfLustreMGT controller to run an `lctl` command on the Rabbit to erase the fsame from the MGT so it can be reused.

For external MGTs, a NnfLustreMGT resource must be created to track fsnames. No lctl commands are run on external MGTs, so the fsname will eventually wrap and cause an error when creating a lustre file system. External MGTs should make use of the FsNameStartReference field in the NnfLustreMGT resource to reference a config map where the next available fsname can be stored. This configmap should not be managed by argocd so that it persists across undeploys/deploys.

The NnfLustreMGT controllers run on both the Rabbits and a worker node. The controllers running on the Rabbits handle NnfLustreMGT resources for internal MGTs where the lctl command can be run. The controller running on the k8s worker handles NnfLustreMGTs that describe external MGTs.


internal/controller/suite_test.go
```diff
@@ -315,16 +315,27 @@ var _ = BeforeSuite(func() {
...
 	err = (&NnfClientMountReconciler{
 		Client:            k8sManager.GetClient(),
 		Log:               ctrl.Log.WithName("controllers").WithName("NnfClientMount"),
 		Scheme:            testEnv.Scheme,
-		SemaphoreForStart: semNnfNodeStorageDone,
+		SemaphoreForStart: semNnfLustreMgtDone,
 	}).SetupWithManager(k8sManager)
 	Expect(err).ToNot(HaveOccurred())
 
+	// The NLC controllers relying on the readiness of EC.
```
roehrich-hpe, line 337: Move this comment above NnfClientMountReconciler.

roehrich-hpe:
What is the lifecycle of that ConfigMap? The admin has to create the initial version of it? How many of these must the admin create? Do they also need to create an NnfLustreMGT for it?

matthew-richerson:
> What is the lifecycle of that ConfigMap? The admin has to create the initial version of it? How many of these must the admin create? Do they also need to create an NnfLustreMGT for it?

An admin creates an empty configmap for each external MGT that is used. This happens once and then the configmap lives forever.
There also needs to be an NnfLustreMGT per external MGT. The NnfLustreMGT can be managed by argocd.

I'm planning on writing up some documentation on how to do that for our docs repo after this PR is done.

cmd/main.go
```diff
@@ -226,17 +226,28 @@ func (c *nodeLocalController) SetupReconcilers(mgr manager.Manager, opts *nnf.Op
...
 		Client:            mgr.GetClient(),
 		Log:               ctrl.Log.WithName("controllers").WithName("NnfClientMount"),
 		Scheme:            mgr.GetScheme(),
-		SemaphoreForStart: semNnfNodeStorageDone,
+		SemaphoreForStart: semNnfLustreMgtDone,
 		ClientType:        controllers.ClientRabbit,
 	}).SetupWithManager(mgr); err != nil {
 		return err
 	}
+	// The NLC controllers relying on the readiness of EC.
```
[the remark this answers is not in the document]
matthew-richerson: I think I'm confused about what these comments mean. I thought everything in between this comment and the comment on line 181 were being started sequentially with the semaphores in the reconciler structs. The LustreMGT controller and the ClientMount controller both don't talk to nnf-ec. Are the semaphores there to coordinate something with nnf-ec? Can the LustreMGT and clientmount controllers start at any time because they don't rely on nnf-ec?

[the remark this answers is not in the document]
roehrich-hpe: NnfClientMount reconciler must wait for EC to be ready, otherwise it will not find storage to mount, even if there are resources telling it to do the mount. That's what the comment is saying.

If NnfLustreMGT can do its stuff without waiting for EC to become ready then it can go where you have it, or after NnfClientMount. If it can do its stuff regardless of the state of EC, then maybe it can be the first reconciler to be started at the top of the list, and it wouldn't need to participate in the semaphore thing and wouldn't need a Start() method.

roehrich-hpe:
> There also needs to be an NnfLustreMGT per external MGT. The NnfLustreMGT can be managed by argocd.

Is that NnfLustreMGT being updated by the reconcilers?

Can that NnfLustreMGT be created by the reconciler on demand?

matthew-richerson:
> > There also needs to be an NnfLustreMGT per external MGT. The NnfLustreMGT can be managed by argocd.
> 
> Is that NnfLustreMGT being updated by the reconcilers?
> 
> Can that NnfLustreMGT be created by the reconciler on demand?

Yes, the NnfLustreMGT gets updated by the NnfStorage reconciler to add a claim to it whenever a new Lustre file system wants to make use of the MGT. And then the NnfLustreMGT reconciler updates the status section with the fsname for that file system to use as well as updating the configmap to reflect the next available fsname. 

Any MGT that's created on a Rabbit node will have a NnfLustreMGT resource created automatically for it by the reconcilers. It's only for an external MGT that an admin would need to create one manually. In theory we could create one automatically for an external MGT, but because that MGT might be part of a Lustre file system that's important, I think it should be left up to an admin (through argocd) to create the resource so they make sure the blacklist and starting fsname is filled in properly. That will let them protect any important file systems served by the MGT and have a record of which fsnames need to eventually be manually deleted off of the MGT.

External MGTs are falling out of favor, though. They're being replaced by MGT pools on Rabbits nodes.

config/samples/nnf_v1alpha1_nnflustremgt.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: nnf.cray.hpe.com/v1alpha1
+kind: NnfLustreMGT
+metadata:
+  labels:
+    app.kubernetes.io/name: nnf-sos
+    app.kubernetes.io/managed-by: kustomize
+  name: nnflustremgt-sample
+spec:
+  # TODO(user): Add fields here
```
bdevcich, line 9: Do you plan on filling this out? If not, then I'd say just remove this file.

internal/controller/integration_test.go
```diff
@@ -1457,13 +1458,27 @@ var _ = Describe("Integration Test", func() {
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "profile-mgs",
+					Namespace: corev1.NamespaceDefault,
+				},
+				Spec: nnfv1alpha1.NnfLustreMGTSpec{
+					Addresses:   []string{profileMgsNid},
+					FsNameStart: "dddddddd",
+				},
+			}
+			Expect(k8sClient.Create(context.TODO(), nnfLustreMgt)).To(Succeed())
```
bdevcich, line 1473: Is just testing create/delete sufficient here?

internal/controller/nnf_storage_controller.go
```diff
@@ -575,6 +622,11 @@ func (r *NnfStorageReconciler) aggregateNodeStorageStatus(ctx context.Context, s
 	}
 
 	for _, nnfNodeStorage := range nnfNodeStorageList.Items {
+		// If we're in the delete path, only propogate errors for storages that are deleting. Errors
```
bdevcich, line 625:
```suggestion
		// If we're in the delete path, only propagate errors for storages that are deleting. Errors
```

internal/controller/nnf_storage_controller.go
```diff
@@ -454,41 +457,85 @@ func (r *NnfStorageReconciler) aggregateNodeBlockStorageStatus(ctx context.Conte
...
+		if allocationSet.TargetType == "mgt" || allocationSet.TargetType == "mgtmdt" {
+			nnfLustreMgt := &nnfv1alpha1.NnfLustreMGT{
 				ObjectMeta: metav1.ObjectMeta{
-					Name:      "nnf-nlc",
+					Name:      storage.GetName(),
 					Namespace: mgsNode,
 				},
+				Spec: nnfv1alpha1.NnfLustreMGTSpec{
+					Addresses:   []string{mgsAddress},
+					FsNameStart: "aaaaaaaa",
```
bdevcich, line 509: Why `aaaaaaaa`?

internal/controller/nnf_storage_controller.go
```diff
@@ -758,20 +913,115 @@ func (r *NnfStorageReconciler) setLustreOwnerGroup(ctx context.Context, nnfStora
...
+	listOptions := []client.ListOption{
+		matchLabels,
+	}
+
+	if err := r.List(ctx, clientMountList, listOptions...); err != nil {
+		return dwsv1alpha2.NewResourceError("could not list ClientMounts").WithError(err)
+	}
+
+	for _, clientMount := range clientMountList.Items {
+		// If we're in the delete path, only propogate errors for storages that are deleting. Errors
```
bdevcich, line 931:
```suggestion
		// If we're in the delete path, only propagate errors for storages that are deleting. Errors
```

config/samples/nnf_v1alpha1_nnflustremgt.yaml
matthew-richerson: No, we don't use these for anything. The ones that are filled out are all stale. I'll just delete it

internal/controller/nnf_lustre_mgt_controller.go
```diff
@@ -0,0 +1,442 @@
...
+	if digit > end {
+		digit = start
+		return append(incrementRuneList(digits[:len(digits)-1], start, end), digit)
+	}
+	digits[len(digits)-1] = digit
+
+	return digits
+}
+
+func incrementFsName(nnfLustreMgt *nnfv1alpha1.NnfLustreMGT, fsname string) string {
```
bdevcich, line 260: `nnfLustreMgt` is not used here

internal/controller/nnf_lustre_mgt_controller.go
```diff
@@ -0,0 +1,442 @@
...
+	var runeList []rune
+	for _, runeDigit := range fsname {
+		runeList = append(runeList, runeDigit)
+	}
+
+	return string(incrementRuneList(runeList, 'a', 'z'))
+}
+
+func isFsNameBlackListed(ctx context.Context, nnfLustreMgt *nnfv1alpha1.NnfLustreMGT, fsname string) bool {
```
bdevcich, line 269: `ctx` is not used

internal/controller/integration_test.go
matthew-richerson: It gets used when a Lustre file system is created, so it's doing more than just a create/delete. I can make some tests for the NnfLustreMGT controller specifically.l

internal/controller/nnf_storage_controller.go
matthew-richerson: The fsnames just increment (e.g., `aaaaaaaa`, `aaaaaaab`, `aaaaaaac`). Starting at `aaaaaaaa` was an arbitrary choice since it's easy to recognize that it's the start of the sequence and that `zzzzzzzz` is the end. Do you have another suggestion?

internal/controller/nnf_storage_controller.go
matthew-richerson: Some day I'll learn how to spell

internal/controller/nnf_storage_controller.go
bdevcich: Nope. Makes sense.

bdevcich (approved):
Nice work!

internal/controller/nnf_lustre_mgt_controller.go
```diff
@@ -0,0 +1,393 @@
...
+
+			return ctrl.Result{Requeue: true}, nil
+		}
+
+		return ctrl.Result{}, nil
+	}
+
+	// Add the finalizer
+	if !controllerutil.ContainsFinalizer(nnfLustreMgt, finalizerNnfLustreMGT) {
+		controllerutil.AddFinalizer(nnfLustreMgt, finalizerNnfLustreMGT)
```
roehrich-hpe, line 120: Maybe we don't have a reason to put a finalizer on this resource?

internal/controller/nnf_lustre_mgt_controller.go
```diff
@@ -0,0 +1,393 @@
...
+
+	return nil
+}
+
+func (r *NnfLustreMGTReconciler) EraseOldFsName(nnfLustreMgt *nnfv1alpha1.NnfLustreMGT, fsname string) error {
+	log := r.Log.WithValues("NnfLustreMGT", client.ObjectKeyFromObject(nnfLustreMgt))
+
+	if r.ControllerType == ControllerRabbit {
+		if cmd, found := os.LookupEnv("NNF_LUSTRE_FSNAME_ERASE_COMMAND"); found {
+			log.Info("Erasing fsname from MGT", "fsname", fsname)
```
roehrich-hpe, line 345: We seem to be conflating MGT with MGS, but it probably doesn't matter.
It occurs to me that almost every use of MGT in this PR is really referring to an MGS rather than an individual MGT.

internal/controller/nnf_lustre_mgt_controller.go
matthew-richerson: I'm planning on adding in some logic on what to do if someone tries to remove the NnfLustreMGT resource while there are still claims on it, so I'd like to leave this in for now.

internal/controller/nnf_lustre_mgt_controller.go
matthew-richerson: I think MGT makes sense here. That's where all of the fsname data is being stored. There can be multiple MGSs that have access to the MGT, and we wouldn't want to track the MGSs individually since they serve up the same target.

api/v1alpha1/nnf_lustre_mgt.go
```diff
@@ -0,0 +1,103 @@
...
+// NnfLustreMGTSpec defines the desired state of NnfLustreMGT
+type NnfLustreMGTSpec struct {
+	// Addresses is the list of LNet addresses for the MGT
+	Addresses []string `json:"addresses"`
+
+	// FsNameBlackList is a list of fsnames that can't be used. This may be
+	// necessary if the MGT hosts file systems external to Rabbit
+	FsNameBlackList []string `json:"fsNameBlackList,omitempty"`
+
+	// FsNameStart is the starting fsname to be used
```
roehrich-hpe, line 39: Use +kubebuilder:validation:MaxLength
https://book.kubebuilder.io/reference/markers/crd-validation

And fit it into other related places.

roehrich-hpe (approved):

roehrich-hpe:
Matt, can you update this with master? It's based on the repo as it was in early June.

PR #342, opened 2024-07-31
Add NnfDataMovementProfiles

bdevcich:
nnf-dm uses a configmap to store its configuration in. This is not consistent with the other means of configuration (i.e. Storage and Container Profiles).

- Converted nnf-dm-config configmap to NnfDataMovementProfile
- NnfDataMovementProfiles live in the nnf-system namespace vs the nnf-dm-system namespace with the configmap
- NnfDataMovementProfiles follow the same behavior as NnfStorageProfiles: pinned profiles are created during Proposal
- Pinned profiles are only created when copy_in/copy_out directives are present
- Data Movement initiated via the Copy Offload API do not use pinned profiles since the user can select any profile they wish and can change profiles with each request
- NnfDataMovements now use a ProfileReference rather than just a string profile name to track namespace/name

Webhook to come afterwards.


config/examples/nnf_v1alpha1_nnfdatamovementprofile.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: nnf.cray.hpe.com/v1alpha1
+kind: NnfDataMovementProfile
+metadata:
+  name: default
```
roehrich-hpe, line 4: Call this one "template" so it doesn't collide with the site-customized default profile.

config/examples/nnf_v1alpha1_nnfdatamovementprofile.yaml
```diff
@@ -0,0 +1,21 @@
...
+  maxSlots: 0
+  command: ulimit -n 2048 && mpirun --allow-run-as-root --hostfile $HOSTFILE dcp --progress 1 --uid $UID --gid $GID $SRC $DEST
+  logStdout: false
+  storeStdout: false
+  progressIntervalSeconds: 5
+---
+apiVersion: nnf.cray.hpe.com/v1alpha1
+kind: NnfDataMovementProfile
+metadata:
+  name: no-xattr
```
roehrich-hpe, line 18: Is the no-xattr profile specifically for tests, or do you consider this to be a production profile?

internal/controller/nnf_workflow_controller_test.go
```diff
@@ -623,6 +835,8 @@ var _ = Describe("NNF Workflow Unit Tests", func() {
 				deletePersistentStorageInstance(persistentStorageName)
 
 				Expect(k8sClient.Delete(context.TODO(), ns)).Should(Succeed())
+
+				Expect(k8sClient.Delete(context.TODO(), dmProfile)).Should(Succeed())
```
roehrich-hpe, line 839: Is the 'ns' in the previous line the same namespace that this 'dmProfile' is in? Do these two need to be swapped?

config/examples/nnf_v1alpha1_nnfdatamovementprofile.yaml
bdevcich: It's used only for copy_out to global lustre, since our global lustre servers do not support extended attributes. copy_in doesn't need this.

internal/controller/nnf_workflow_controller_test.go
bdevcich: Yes. I'll swap them because it makes sense. However, envtest doesn't [support deleting a namespace](https://github.com/kubernetes-sigs/controller-runtime/issues/880), so it probably doesn't matter until it gets fixed.

api/v1alpha1/workflow_helpers.go
```diff
@@ -50,6 +50,14 @@ const (
 	// which pinned container profile is being used.
 	PinnedContainerProfileLabelNameSpace = "nnf.cray.hpe.com/pinned_container_profile_namespace"
 
+	// PinnedDataMovementProfileLabelName is a label applied to NnfStorage objects to show
```
matthew-richerson, line 53: Should this be NnfDataMovement?

config/examples/nnf_v1alpha1_nnfdatamovementprofile.yaml
bdevcich: Do I set `default:false` too then?

api/v1alpha1/workflow_helpers.go
bdevcich: The other profile types (i.e. storage, container) aren't using `Nnf`, so I followed suit here.

matthew-richerson: I meant s/NnfStorage/NnfDataMovement/

config/examples/nnf_v1alpha1_nnfdatamovementprofile.yaml
roehrich-hpe: yes

roehrich-hpe: The unpack-manifest tool will use this template to the create the default resource, and it will set default=true when it does that.

config/examples/nnf_v1alpha1_nnfdatamovementprofile.yaml
roehrich-hpe: Then this is site-specific, so it should live only in the gitops repo, not here in the nnf-sos repo.

api/v1alpha1/workflow_helpers.go
bdevcich: Oh whoops. I'll fix that.

bdevcich: Actually, these aren't even used since I've updated NnfDatamovement to use ObjectReferences to point the profile being used.

I think I can remove these altogether.

roehrich-hpe (approved):

matthew-richerson (approved):

PR #345, opened 2024-08-02
Add environment variables to set QPS and burst values

matthew-richerson:
The rest.Config QPS and burst values control client side throttling. Expose these values through environment variables so they can be tuned.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -23,6 +23,7 @@ import (
 	"flag"
 	"os"
 	"runtime"
+	"strconv"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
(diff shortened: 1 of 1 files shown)


bdevcich:
We set these values in the nnf-dm daemon. Is that affected by this or only clientmountd?

https://github.com/NearNodeFlash/nnf-dm/blob/master/daemons/compute/server/servers/server.go#L54-L58

bdevcich (approved):

matthew-richerson:
> We set these values in the nnf-dm daemon. Is that affected by this or only clientmountd?
> 
> https://github.com/NearNodeFlash/nnf-dm/blob/master/daemons/compute/server/servers/server.go#L54-L58

Oh cool, I didn't realize nnf-dm was changing those values too. This PR is for the nnf-controller-manager. I think we'll want to set the values pretty high (I tested with QPS=500 and Burst=1000) since it's only for a single client/pod, that pod has a large number of reconcilers in it, and it's doing a bunch of scale out operations for the ClientMount, NnfNodeStorage, and NnfNodeBlockStorage resources.

The nnf-dm daemon and clientmountd are only affected because the API server will have more traffic from the nnf-controller-manager.

bdevcich:
Thanks. I saw "client" and made a poor assumption that it would effect clientmount.

For nnf-dm, I just picked those values somewhat arbitrarily, but they were performing much better than the default.

matthew-richerson:
> Thanks. I saw "client" and made a poor assumption that it would effect clientmount.
> 
> For nnf-dm, I just picked those values somewhat arbitrarily, but they were performing much better than the default.

The values for client-go (which nnf-dm is using) are really low (QPS=5, Burst=10). The controller-runtime client used by nnf-controller-manager has QPS=20, Burst=30 by default. We'll have to see what the API server can handle. We'll probably have to make those values configurable in clientmountd as well.

roehrich-hpe (approved):

PR #347, opened 2024-08-03
Check MGT status when picking from a pool

matthew-richerson:
Check the status of the NnfStorage and the DWS Storage resource before picking an MGT from a pool. This should decrease the possibility that a file system tries to use an MGT that isn't functional.


internal/controller/nnf_workflow_controller_helpers.go
```diff
@@ -713,37 +713,86 @@ func (r *NnfWorkflowReconciler) getLustreMgsFromPool(ctx context.Context, pool s
 		return corev1.ObjectReference{}, "", dwsv1alpha2.NewResourceError("").WithUserMessage("no MGSs found for pool: %s", pool).WithFatal().WithUser()
 	}
 
-	persistentStorage := persistentStorageList.Items[rand.Intn(len(persistentStorageList.Items))]
+	healthyMgts := make(map[string]corev1.ObjectReference)
+	for _, persistentStorage := range persistentStorageList.Items {
+		// Find the NnfStorage for the PersistentStorage so we can check it's status and get the MGT LNid
```
roehrich-hpe, line 718:
```suggestion
		// Find the NnfStorage for the PersistentStorage so we can check its status and get the MGT LNid
```

internal/controller/nnf_workflow_controller_helpers.go
```diff
@@ -713,37 +713,86 @@ func (r *NnfWorkflowReconciler) getLustreMgsFromPool(ctx context.Context, pool s
...
-	}
+		if err := r.Get(ctx, client.ObjectKeyFromObject(nnfStorage), nnfStorage); err != nil {
+			return corev1.ObjectReference{}, "", dwsv1alpha2.NewResourceError("could not get persistent NnfStorage %v for MGS", client.ObjectKeyFromObject(nnfStorage)).WithError(err)
+		}
 
-	if err := r.Get(ctx, client.ObjectKeyFromObject(nnfStorage), nnfStorage); err != nil {
-		return corev1.ObjectReference{}, "", dwsv1alpha2.NewResourceError("could not get persistent NnfStorage %v for MGS", client.ObjectKeyFromObject(nnfStorage)).WithError(err)
-	}
+		// Do some sanity checks on the NnfStorage to make sure it's really a standalone MGT
+		if nnfStorage.Spec.FileSystemType != "lustre" {
```
roehrich-hpe, line 731: Do we have some other place where these sanity checks happen?
I think you should log it, and then go to the next one rather than returning Fatal.
Maybe check nnfStorage.Status.Ready first, rather than last.

internal/controller/nnf_workflow_controller_helpers.go
```diff
@@ -713,37 +713,86 @@ func (r *NnfWorkflowReconciler) getLustreMgsFromPool(ctx context.Context, pool s
...
 
-	// Find the NnfStorage for the PersistentStorage so we can get the LNid
-	nnfStorage := &nnfv1alpha1.NnfStorage{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      persistentStorage.Name,
-			Namespace: persistentStorage.Namespace,
-		},
-	}
+		if err := r.Get(ctx, client.ObjectKeyFromObject(nnfStorage), nnfStorage); err != nil {
+			return corev1.ObjectReference{}, "", dwsv1alpha2.NewResourceError("could not get persistent NnfStorage %v for MGS", client.ObjectKeyFromObject(nnfStorage)).WithError(err)
```
roehrich-hpe, line 727: Log it, and go to the next one.

internal/controller/nnf_workflow_controller_helpers.go
```diff
@@ -713,37 +713,86 @@ func (r *NnfWorkflowReconciler) getLustreMgsFromPool(ctx context.Context, pool s
...
+	// Check to make sure there's at least one MGT we can use
+	if len(healthyMgts) == 0 {
+		return corev1.ObjectReference{}, "", dwsv1alpha2.NewResourceError("").WithUserMessage("no healthy MGSs found for pool: %s", pool).WithMajor()
 	}
 
-	if len(nnfStorage.Status.MgsAddress) == 0 {
-		return corev1.ObjectReference{}, "", dwsv1alpha2.NewResourceError("no LNid listed for persistent MGS").WithFatal()
+	// Choose an MGT at random from the map
+	i := rand.Intn(len(healthyMgts))
+	for mgsAddress, persistentStorageReference := range healthyMgts {
```
roehrich-hpe, line 788: This sent me off to do some reading, and wow, Go sucks here, doesn't it? It looks like some reasonable help is coming in Go 1.23, but until then you're stuck using "experimental" libraries to make this better.

internal/controller/nnf_workflow_controller_helpers.go
matthew-richerson: If these checks fail, then something is seriously wrong. These are the only values that the nnfstorage can have if it was set up as a Lustre MGT. If the values don't match, then someone must have created an nnfstorage manually.

internal/controller/nnf_workflow_controller_helpers.go
matthew-richerson: I'd rather not do that in case it's the only MGT that's available. If the cache is stale we might get this error, but it's marked as a Minor() error (the default), so it never even makes it out to flux anyway. It just requeues the reconciler to try again.

internal/controller/nnf_workflow_controller_helpers.go
matthew-richerson: Yeah, I was disappointed in this too.

roehrich-hpe (approved):

PR #365, opened 2024-08-29
Add NnfSystemStorage for manual storage creation

matthew-richerson:
The NnfSystemStorage resource is used for creating storage outside of a workflow. The main use case for this is a pair of long lived LVM volume groups that will be activated on the compute nodes as a workaround for an lvmlockd bug.


api/v1alpha1/nnfsystemstorage_types.go
```diff
@@ -0,0 +1,136 @@
...
+	// +kubebuilder:validation:items:Maximum=15
+	// +kubebuilder:validation:items:Minimum=0
+	ComputesPattern []int `json:"computesPattern,omitempty"`
+
+	// Capacity is the allocation size on each Rabbit
+	// +kubebuilder:default:=1073741824
+	Capacity int64 `json:"capacity"`
+
+	// Type is the file system type to use for the storage allocation
+	// +kubebuilder:validation:Enum=raw
```
roehrich-hpe, line 77: Let's add XFS right away. Then you don't have to fuss with it down the road when you would have to make a conversion routine for it.

roehrich-hpe (approved):

api/v1alpha1/nnfsystemstorage_types.go
matthew-richerson: Okay. I'll do gfs2 as well then.

api/v1alpha1/nnfsystemstorage_types.go
```diff
@@ -0,0 +1,136 @@
...
+	// SystemConfiguration
+	IncludeRabbits []string `json:"includeRabbits,omitempty"`
+
+	// ExcludeComputes is a list of compute nodes to exclude from the the compute nodes listed in the
+	// SystemConfiguration
+	ExcludeComputes []string `json:"excludeComputes,omitempty"`
+
+	// IncludeComputes is a list of computes nodes to use rather than getting the list of compute nodes
+	// from the SystemConfiguration
+	IncludeComputes []string `json:"includeComputes,omitempty"`
```
bdevcich, line 58: By default, does this just grab random rabbits and computes from the SystemConfiguration if you don't use these include/exclude fields?

internal/controller/nnfsystemstorage_controller.go
```diff
@@ -0,0 +1,632 @@
+/*
```
bdevcich, line 1: The other controllers have an underscore after nnf in the filename.

I'm guilty of this too, since I've added datamovement/container profiles and didn't use an underscore, but it's bothered me.

internal/controller/nnfsystemstorage_controller.go
```diff
@@ -0,0 +1,632 @@
...
+func (r *NnfSystemStorageReconciler) createServers(ctx context.Context, nnfSystemStorage *nnfv1alpha1.NnfSystemStorage) error {
+	log := r.Log.WithValues("NnfSystemStorage", client.ObjectKeyFromObject(nnfSystemStorage))
+
+	// Create a list of Rabbits to use
+	rabbitList := []string{}
+
+	if len(nnfSystemStorage.Spec.IncludeRabbits) != 0 {
+		if len(nnfSystemStorage.Spec.ExcludeRabbits) != 0 {
+			return dwsv1alpha2.NewResourceError("IncludeRabbits and ExcludeRabbits can not be used together").WithFatal()
+		}
```
bdevcich, line 227: Should this be in a webhook?

internal/controller/nnfsystemstorage_controller.go
matthew-richerson: It bothers me too, but I wonder if we should switch them all to use no underscore? That's the name of the file as generated by kubebuilder. I think kubebuilder must have changed at some point over the years.

I'm open to suggestions.

bdevcich: Ah, yes. I think we should change them all to be consistent and we should go with what Kubebuilder does. Probably outside the scope of this PR though, but if you're feeling motivated :wink:

bdevcich (approved):

api/v1alpha1/nnfsystemstorage_types.go
bdevcich: Makes sense. I think you should make sure to document that somewhere. If not here, then definitely in the nnf docs.

internal/controller/nnfsystemstorage_controller.go
matthew-richerson: I like the idea, but it's probably not that useful for a resource that is going to see such little use. Maybe I can add it as a follow on so we can get the base functionality out in a release sooner?

api/v1alpha1/nnfsystemstorage_types.go
matthew-richerson: Yeah, I'm going to do a write up on how to use this in the nnf docs.

PR #377, opened 2024-09-05
Allow empty FStype in NnfNodeStorageSpec

roehrich-hpe:
For the kubebuilder:default to work, you have to first allow the field to be empty.


matthew-richerson (approved):

matthew-richerson:
What's the reason for needing this?

roehrich-hpe:
> What's the reason for needing this?

Just stuff I noticed while enabling the tests for the conversion webhooks.

PR #382, opened 2024-09-09
Add deleteUnknownVolumes cli parameter to nnf-ec

ajfloeder:
Vendor the latest nnf-ec and supply the `deleteUnknownVolumes` cli parameter.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -220,6 +220,7 @@ func (c *nodeLocalController) SetupReconcilers(mgr manager.Manager, opts *nnf.Op
 		Events:           make(chan event.GenericEvent),
 		SemaphoreForDone: semNnfNodeDone,
 		NamespacedName:   types.NamespacedName{Name: controllers.NnfNlcResourceName, Namespace: os.Getenv("NNF_NODE_NAME")},
+		Options:          opts,
 	}).SetupWithManager(mgr); err != nil {
 		return err
 	}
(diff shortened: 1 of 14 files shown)


roehrich-hpe (approved):

config/default/manager_auth_proxy_patch.yaml
```diff
@@ -25,3 +25,4 @@ spec:
         - "--metrics-bind-address=127.0.0.1:8080"
         - "--leader-elect"
         - "--controller=storage"
+        - "--deleteUnknownVolumes=true"
```
matthew-richerson, line 28: does this value get passed to nnf-ec?

ajfloeder: I have to test it, but I believe so.

roehrich-hpe: Probably want to be here, in the nnf-node-manager daemonset:
https://github.com/NearNodeFlash/nnf-sos/blob/master/config/dp0/manager_volumes_patch.yaml#L12

That just happens to be place where we set the Daemonset's commandline args for real hardware.
I guess we wouldn't need need this in a KIND env, right?

matthew-richerson: I think Dean has the correct place to put the new arg. Looking at the pods on the system, the pod running on the k8s worker node has "deleteUnknownVolumes=true", and the pods running on the Rabbits don't:
```
matthew.richerson@HPE-TV66XWF7QG nnf-sos % kubectl get pods -n nnf-system nnf-node-manager-f95rr -o yaml
apiVersion: v1
kind: Pod
...
  containers:
  - args:
    - --controller=node
    command:
    - /manager
    ...
```

roehrich-hpe (approved):

matthew-richerson (approved):

PR #388, opened 2024-09-11
Trigger reconcile for NnfNodeBlockStorage from nnf-ec events

matthew-richerson:
Use the GenericEvent source to trigger the event through a channel. Update the NnfNode reconciler to use the same method.

Also, change the NnfAccess reconciler to wait for the StorageGroups to be created on the Rabbit before creating the ClientMount resources.

### DIFF
--- a/internal/controller/nnf_access_controller.go
+++ b/internal/controller/nnf_access_controller.go
@@ -212,7 +212,7 @@ func (r *NnfAccessReconciler) mount(ctx context.Context, access *nnfv1alpha2.Nnf
 		return &ctrl.Result{RequeueAfter: time.Second}, nil
 	}
 
-	// Add compute node information to the storage map, if necessary.
+	// Request that the devices be made available on the correct nodes
 	err = r.addBlockStorageAccess(ctx, access, storageMapping)
 	if err != nil {
 		if apierrors.IsConflict(err) {
@@ -220,6 +220,16 @@ func (r *NnfAccessReconciler) mount(ctx context.Context, access *nnfv1alpha2.Nnf
 		}
 
 		return nil, dwsv1alpha2.NewResourceError("unable to add endpoints to NnfNodeStorage").WithError(err)
+	}
+
+	// Wait for all the devices to be made available on the correct nodes
+	ready, err := r.getBlockStorageAccessStatus(ctx, access, storageMapping)
+	if err != nil {
+		return nil, dwsv1alpha2.NewResourceError("unable to check endpoints for NnfNodeStorage").WithError(err)
+	}
+
+	if ready == false {
+		return &ctrl.Result{RequeueAfter: time.Second * 2}, nil
 	}
 
 	// Create the ClientMount resources. One ClientMount resource is created per client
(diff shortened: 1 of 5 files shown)


internal/controller/nnf_node_block_storage_controller.go
```diff
@@ -587,14 +629,57 @@ func (r *NnfNodeBlockStorageReconciler) deleteStorageGroup(ss nnf.StorageService
...
+	requests := []reconcile.Request{}
+
+	// Find all the NnfNodeBlockStorage resources for this Rabbit so we can reconcile them.
+	listOptions := []client.ListOption{
+		client.InNamespace(os.Getenv("NNF_NODE_NAME")),
+	}
+
+	nnfNodeBlockStorageList := &nnfv1alpha2.NnfNodeBlockStorageList{}
+	if err := r.List(context.TODO(), nnfNodeBlockStorageList, listOptions...); err != nil {
+		log.Error(err, "Could not list block storages")
```
roehrich-hpe, line 646: How common will this be? Do we really want the backtrace?

internal/controller/nnf_access_controller.go
```diff
@@ -222,6 +222,16 @@ func (r *NnfAccessReconciler) mount(ctx context.Context, access *nnfv1alpha2.Nnf
...
 	}
 
+	// Wait for all the devices to be made available on the correct nodes
+	ready, err := r.getBlockStorageAccessStatus(ctx, access, storageMapping)
+	if err != nil {
+		return nil, dwsv1alpha2.NewResourceError("unable to check endpoints for NnfNodeStorage").WithError(err)
+	}
+
+	if ready == false {
+		return &ctrl.Result{RequeueAfter: time.Second * 2}, nil
```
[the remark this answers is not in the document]
roehrich-hpe: What's the alternative?
I think we'd better just keep waiting until the pod is killed.

internal/controller/nnf_node_block_storage_controller.go
matthew-richerson: These events aren't that common, and a List() failing isn't very common, but if you're worried I can change it to an Info() with the error message included.

roehrich-hpe (approved):

internal/controller/nnf_node_block_storage_controller.go
```diff
@@ -72,13 +78,48 @@ type NnfNodeBlockStorageReconciler struct {
...
 
+// EventHandler implements event.Subscription. Every Upstream or Downstream event triggers a watch
+// on all the NnfNodeBlockStorages. This is needed to create the StorageGroup for a compute node that
+// was powered off when the Access list was updated.
+func (r *NnfNodeBlockStorageReconciler) EventHandler(e nnfevent.Event) error {
+	log := r.Log.WithValues("nnf-ec event", "node-up/node-down")
+
+	// Upstream link events
+	upstreamLinkEstablished := e.Is(msgreg.UpstreamLinkEstablishedFabric("", "")) || e.Is(msgreg.DegradedUpstreamLinkEstablishedFabric("", ""))
+	upstreamLinkDropped := e.Is(msgreg.UpstreamLinkDroppedFabric("", ""))
```
bdevcich, line 94: are these nnf-ec events then?

matthew-richerson: Yep

PR #389, opened 2024-09-13
Ignore disabled Rabbits in NnfSystemStorage

matthew-richerson:
If a Rabbit is marked as "Disabled" in the DWS storage resource, add an option to exlude it from the NnfSystemStorage Rabbit list.

Also, the Storage resource is created in DWS now, so remove the creation in nnf-sos. Add Watch()'s for various resources so that the NnfSystemStorage automatically adds storage on Rabbits that transition from "Disabled" to "Enabled".


internal/controller/nnfsystemstorage_controller.go
```diff
@@ -254,6 +257,40 @@ func (r *NnfSystemStorageReconciler) createServers(ctx context.Context, nnfSyste
...
+			labels := storage.GetLabels()
+			if labels == nil {
+				continue
+			}
+
+			if storageType := labels[dwsv1alpha2.StorageTypeLabel]; storageType != "Rabbit" {
+				continue
+			}
+
+			if storage.Spec.State == dwsv1alpha2.DisabledState || storage.Status.Status == dwsv1alpha2.DisabledStatus {
```
roehrich-hpe, line 285: Should we instead skip anything that has a status != ReadyStatus?

ajfloeder (approved):

roehrich-hpe (approved):

api/v1alpha1/conversion.go
```diff
@@ -21,6 +21,7 @@ package v1alpha1
 
 import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	apiconversion "k8s.io/apimachinery/pkg/conversion"
```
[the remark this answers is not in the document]
matthew-richerson: That's a different conversion.go file. I just copied from the dws conversion.go file.

internal/controller/nnf_access_controller.go
```diff
@@ -1086,6 +1148,7 @@ func (r *NnfAccessReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		WithOptions(controller.Options{MaxConcurrentReconciles: maxReconciles}).
 		For(&nnfv1alpha2.NnfAccess{}).
+		Watches(&dwsv1alpha2.Computes{}, handler.EnqueueRequestsFromMapFunc(r.ComputesEnqueueRequests)).
```
bdevcich, line 1151: so this affectively triggers a reconcile with the NnfAccess controller?

matthew-richerson: Yes, if the Computes resource for an NnfSystemStorage (or a Workflow) changes, the associated NnfAccess will reconcile again. For the Workflow, the Computes resource won't change, so this really only affects NnfSystemStorage.

bdevcich (approved):

PR #410, opened 2024-10-29
Only aggregate status from children that would be created using the c…

matthew-richerson:
…urrent spec

For NnfStorage and NnfAccess resources created by the NnfSystemStorage, the spec section may change as Storage resources are disabled/enabled. When aggregating status from child objects (NnfNodeBlockStorage, NnfNodeStorage, and ClientMounts), only check the status from child resources that are currently requested by the spec. This avoids trying to collect status from Rabbits that are disabled.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/NearNodeFlash/nnf-sos
 go 1.21
 
 require (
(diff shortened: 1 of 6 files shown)


roehrich-hpe (approved):

mcfadden8:
just curious - will this fix address symptoms that we might have seen previously?

ajfloeder (approved):

PR #412, opened 2024-11-04
DRAFT: Post Mount Logic Check

bdevcich:
The bits in `nnf_node_storage_controller.go` is the main part.

### DIFF
--- a/internal/controller/nnf_node_storage_controller.go
+++ b/internal/controller/nnf_node_storage_controller.go
@@ -22,7 +22,6 @@ package controller
 import (
 	"context"
 	"runtime"
-	"strconv"
 	"sync"
 	"time"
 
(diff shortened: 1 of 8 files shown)


internal/controller/nnf_node_storage_controller.go
```diff
@@ -363,12 +364,86 @@ func (r *NnfNodeStorageReconciler) createAllocations(ctx context.Context, nnfNod
...
+
 		allocationStatus.Ready = true
 	}
 
 	return nil, nil
 }
 
+// For a given NnfNodeStorage, retrieve all the associated NnfNodeBlockStorages and check if they
+// are all Ready. Obtain the list of NnfNodeBlockStorages by the NnfStorage owner.
+func (r *NnfNodeStorageReconciler) checkForBlockStorageReady(ctx context.Context, nnfNodeStorage *nnfv1alpha3.NnfNodeStorage) (bool, error) {
```
matthew-richerson, line 406: I must have mistyped something when describing this.

We won't create any NnfNodeStorages until all of the NnfNodeBlockStorages are ready. You just need to make sure that all of the NnfNodeStorages are ready. There should be one NnfNodeStorage for each Rabbit/target type.

The NnfNodeBlockStorages correspond to all the steps up to making /dev/nvme[*]n[*] devices on the Rabbit.
For Lustre, the NnfNodeStorages correspond to combining those NVMe namespaces with a zpool, and mounting a Lustre target on top (e.g., an OST).

Before making a Lustre client mount we want to make sure that all the Lustre targets are mounted.

pkg/filesystem/simple.go
```diff
@@ -275,5 +276,13 @@ func (f *SimpleFileSystem) PreDeactivate(ctx context.Context) (bool, error) {
 		return false, fmt.Errorf("could not unmount after pre-deactivate '%s': %w", f.TempDir, err)
 	}
 
-	return false, nil
+	return true, nil
+}
+
+func (f *SimpleFileSystem) PostMount(ctx context.Context, complete bool) (bool, error) {
+
+	// TODO: Make this the same as PostActivate? Rename PostActivate Here? Have them both do the same thing?
```
matthew-richerson, line 284: Yeah, just rename PostActivate to PostMount. PostActivate will only be a Lustre thing. Or I guess PostActivate could be for XFS and GFS2 as well. It would just allow you to run commands after the mkfs but before a mount of the file system. I don't think that's useful though.

bdevcich:
From @matthew-richerson:

> I just realized while reading through your PR that this isn't going to work
> 
> the nnf-node-manager where that controller runs has a client cache that's limited to the namespace of whatever the Rabbit hostname is
> 
> those NnfNodeStorage resources could be spread across lots of different Rabbits, so lots of different namespaces
> I think the NnfStorage controller is going to have to be the one that coordinates this
> 
> an easy way would be to have the NnfStorage controller create the NnfNodeStorage for OST0 only after all the other NnfNodeStorages are Ready
> 
> then the NnfNodeStorage controller doesn't have to do any extra checking, it just runs PostMount() if it's OST0
> 
> we'd lose a little bit of speed setting up Lustre file systems, but most parts would still happen in parallel

pkg/filesystem/simple.go
bdevcich: I'll make PostActivate/Deactive for xfs/gfs2 a no-op.

PR #417, opened 2024-11-20
Add shared option to NnfSystemStorage

matthew-richerson:
The shared option determines whether to create a single allocation per Rabbit that is shared between all the computes on the Rabbit, or one allocation per compute node.


bdevcich:
What's the reason for this change again?

api/v1alpha1/conversion.go
```diff
@@ -486,8 +486,10 @@ func (src *NnfSystemStorage) ConvertTo(dstRaw conversion.Hub) error {
 	// Otherwise, you may comment out UnmarshalData() until it's needed.
 	if hasAnno {
 		dst.Spec.ExcludeDisabledRabbits = restored.Spec.ExcludeDisabledRabbits
+		dst.Spec.Shared = restored.Spec.Shared
 	} else {
 		dst.Spec.ExcludeDisabledRabbits = false
+		dst.Spec.Shared = true
```
bdevcich, line 492: In alpha2/3, you have this set to false, but true here.

matthew-richerson:
> What's the reason for this change again?

So that a XFS system storage can have a separate file system for each compute node.

api/v1alpha4/nnfsystemstorage_types.go
```diff
@@ -83,6 +83,11 @@ type NnfSystemStorageSpec struct {
 	// +kubebuilder:default:=raw
 	Type string `json:"type,omitempty"`
 
+	// Shared will create one allocation per Rabbit rather than one allocation
+	// per compute node.
+	// +kubebuilder:default:=true
+	Shared bool `json:"shared"`
+
 	// StorageProfile is an object reference to the storage profile to use
```
bdevcich, line 91: Where does the user set this? Gitops?

bdevcich (approved):

api/v1alpha4/nnfsystemstorage_types.go
matthew-richerson: Anywhere you can create a NnfSystemStorage resource. kubectl, gitops, etc.

roehrich-hpe (approved):

bdevcich (approved):

matthew-richerson:
rebased on master to resolve conflicts

PR #418, opened 2024-11-20
Fix aggressive lvRemove and vgRemove

bdevcich:
The default storage profile was only using the $VG_NAME, causing
lvRemove to remove all logical volumes in the volume group.
Additionally, volume groups were being removed without first checking to
see if logical volumes exist in the volume group.

While both of these issues have no real affect on the removal of the
LVs/VGs, this was causing issues with the new PreUnmount commands. The
first logical volume to run the lvRemove command was the only
fileysystem to run the PreUnmount commands, since it destroys all the
other filesystems on the rabbit in that single volume group.

With this change, the PreUnmount command runs on each filesystem before
it is destroyed.

### DIFF
--- a/pkg/blockdevice/lvm.go
+++ b/pkg/blockdevice/lvm.go
@@ -140,10 +140,20 @@ func (l *Lvm) Destroy(ctx context.Context) (bool, error) {
 	destroyed, err := l.LogicalVolume.Remove(ctx, l.CommandArgs.LvArgs.Remove)
 	if err != nil {
 		return false, err
-
 	}
 	if destroyed {
 		objectDestroyed = true
+	}
+
+	// Check to ensure the VG has no LVs before removing
+	lvs, err := lvm.LvsListVolumes(ctx, l.VolumeGroup.Log)
+	if err != nil {
+		return false, err
+	}
+	for _, lv := range lvs {
+		if lv.VGName == l.VolumeGroup.Name {
+			return false, nil
+		}
 	}
 
 	destroyed, err = l.VolumeGroup.Remove(ctx, l.CommandArgs.VgArgs.Remove)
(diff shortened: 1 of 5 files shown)


roehrich-hpe (approved):

pkg/blockdevice/lvm.go
```diff
@@ -140,12 +140,22 @@ func (l *Lvm) Destroy(ctx context.Context) (bool, error) {
...
 	}
 
+	// Check to ensure the VG has no LVs before removing
+	lvs, err := lvm.LvsListVolumes(ctx, l.VolumeGroup.Log)
+	if err != nil {
+		return false, err
+	}
+	for _, lv := range lvs {
+		if lv.VGName == l.VolumeGroup.Name {
+			return false, nil
```
matthew-richerson, line 155: Should this be true since it destroyed the LV?

bdevcich: To me, hitting this line means that there is another LV that is still using the VG. This LV is now gone, but there are still others. 

Do I have that wrong?

pkg/blockdevice/lvm/lvs.go
```diff
@@ -43,7 +43,7 @@ type lvsLogicalVolume struct {
 	Size   string `json:"lv_size"`
 }
 
-func lvsListVolumes(ctx context.Context, log logr.Logger) ([]lvsLogicalVolume, error) {
+func LvsListVolumes(ctx context.Context, log logr.Logger) ([]lvsLogicalVolume, error) {
```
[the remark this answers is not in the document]
bdevcich: That's way better/cleaner. I'll do that.

pkg/blockdevice/lvm.go
bdevcich: Ah, I see what you're saying now. I'll change that.

This got me thinking though: if we remove all the LVs and then somehow on the last one we don't remove the VG because `lvs` is stale or something. A. is that possible and B. are we going to leave around a VG in that scenario?

matthew-richerson (approved):

PR #420, opened 2024-11-20
Enable mock filesystems in KIND

roehrich-hpe:
On the docker host, the /tmp/nnf dir is expected to be mounted into each docker container as /mnt/nnf. This should be specified as an extraMounts in the KIND config file.

The k8s pods will add volume mounts for /mnt/nnf.

When using KIND, new mock devices will be represented as directories in /mnt/nnf. New filesystems will be represented as directories in /mnt/nnf, containing per-rabbit symlinks back to the mock "device" directory.

For a user container to use these mock filesystems, their pod must have a volume mount for the mock filesystem and for the mock device directory, where the symlink is pointing.

### DIFF
--- a/internal/controller/nnf_workflow_controller_helpers.go
+++ b/internal/controller/nnf_workflow_controller_helpers.go
@@ -1905,6 +1905,17 @@ func (r *NnfWorkflowReconciler) getContainerJobs(ctx context.Context, workflow *
 	return jobList, nil
 }
 
+func (r *NnfWorkflowReconciler) getNnfNodeStorages(ctx context.Context, workflow *dwsv1alpha2.Workflow) (*nnfv1alpha4.NnfNodeStorageList, error) {
+	matchLabels := dwsv1alpha2.MatchingWorkflow(workflow)
+
+	nodeStorages := &nnfv1alpha4.NnfNodeStorageList{}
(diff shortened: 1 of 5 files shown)


roehrich-hpe:
This requires https://github.com/NearNodeFlash/nnf-deploy/pull/232

internal/controller/nnf_workflow_controller_helpers.go
```diff
@@ -1991,9 +2002,48 @@ func (r *NnfWorkflowReconciler) getContainerVolumes(ctx context.Context, workflo
...
+	volumes := []nnfContainerVolume{}
+
+	// If we're using the KIND mock storage then we also have to create a
+	// volume mount for the path that represents the device beneath the
+	// filesystem.
+	nodeStoragesList, err := r.getNnfNodeStorages(ctx, workflow)
+	if err != nil {
+		return nil, dwsv1alpha2.NewResourceError("could not find devices for KIND environment").WithError(err)
+	}
+	// On GFS2, the same device is visible on multiple rabbits. Track dups and
```
bdevcich, line 2027:
```suggestion
	// On GFS2, the same device is visible on multiple rabbits. Track dupes and
```

bdevcich (approved):

matthew-richerson (approved):

PR #424, opened 2024-12-04
PostMount Env Variables (e.g. NUM_OSTS)

bdevcich:
There is no way to get the number of OSTs, etc when using `PostMount` commands. For instance, when setting the striping.

This change adds the following environment variables for use in the `NnfStorageProfiles` when using `*CmdLines`:

- NUM_MDTS
- NUM_MGTS
- NUM_MGTMDTS
- NUM_OSTS
- NUM_NNFNODES

To support this, a list of nodes for each component type has been added to the status of `NnfStorage`, which in turn is copied to the `NnfNodeStorage` resource's spec. This info can then be turned into the environment variables for use when running the commands.

The `.nnf-servers.json` file created to store this information on the compute node has been updated to this new structure.

An example with 8 OSTs and 2 MGTMDTs on 1 rabbit:

```json
{
  "mdt": [],
  "mgt": [],
  "mgtmdt": [
    "rabbit-node-1",
    "rabbit-node-1"
  ],
  "nnfNode": [
    "rabbit-node-1"
  ],
  "ost": [
    "rabbit-node-1",
    "rabbit-node-1",
    "rabbit-node-1",
    "rabbit-node-1",
    "rabbit-node-1",
    "rabbit-node-1",
    "rabbit-node-1",
    "rabbit-node-1"
  ]
}
```


api/v1alpha2/conversion.go
```diff
@@ -407,12 +415,21 @@ func (src *NnfStorage) ConvertTo(dstRaw conversion.Hub) error {
...
+	hasAnno, err := utilconversion.UnmarshalData(src, restored)
+	if err != nil {
 		return err
 	}
+
 	// EDIT THIS FUNCTION! If the annotation is holding anything that is
 	// hub-specific then copy it into 'dst' from 'restored'.
 	// Otherwise, you may comment out UnmarshalData() until it's needed.
+	if hasAnno {
+		dst.Status.LustreComponents.MDTs = append([]string(nil), restored.Status.LustreComponents.MDTs...)
```
roehrich-hpe, line 427: I would not have expected `(x)` to work for initializing that string array. I would have expected `{x}` instead, but that doesn't work. Huh.

I wonder what that's called.

roehrich-hpe (approved):

api/v1alpha2/conversion.go
bdevcich: Yea, this one really threw me for a loop.

matthew-richerson (approved):

PR #435, opened 2025-01-09
A datamovement profile for the copy-offload server

roehrich-hpe:


config/examples/copyoffload-datamovement-profile.yaml
```diff
@@ -0,0 +1,17 @@
...
+  maxSlots: 0
+  mkdirCommand: mpirun -np 1 --hostfile $HOSTFILE -- mkdir -p $PATH
+  pinned: false
+  progressIntervalSeconds: 5
+  slots: 8
+  statCommand: mpirun -np 1 --hostfile $HOSTFILE -- stat --cached never -c '%F' $PATH
+  storeStdout: true
+kind: NnfDataMovementProfile
+metadata:
+  name: copy-offload-nonmpi
```
bdevcich, line 16: why nonmpi?

bdevcich (approved):

config/examples/copyoffload-datamovement-profile.yaml
roehrich-hpe: Let's let this PR sit while I prepare the PR that has the container profile and then we can see what makes sense.

roehrich-hpe: I'm moving this profile to nnf-dm, in https://github.com/NearNodeFlash/nnf-dm/pull/255
You'll see it's paired with its container profile.

bdevcich (approved):

PR #439, opened 2025-01-29
Add LV tags for compute node hostname

matthew-richerson:
This commit adds a method for providing variable substitution to commands run on the Rabbits
from the upper layers. This is used by the NnfSystemStorage controller to add a variable that
expands to the host name of one compute node per Rabbit allocation. This allows the NnfStorageProfile
used by the NnfSystemStorage to use the $COMPUTE_HOSTNAME variable in the lvcreate command to
add a tag to each LV corresponding to a compute node hostname.


ajfloeder (approved):

roehrich-hpe (approved):
Please use a merge commit on this one.

PR #441, opened 2025-02-05
Rescan NVMe devices in ClientMount reconciler

matthew-richerson:
If the block debice fails to activate in the ClientMount reconciler, rescan for NVMe devices and let the reconciler retry. Some compute nodes have trouble automatically finding the namespaces after a storage group is created on the Rabbit.

### DIFF
--- a/pkg/blockdevice/lvm/logical_volumes.go
+++ b/pkg/blockdevice/lvm/logical_volumes.go
@@ -59,7 +59,7 @@ func (lv *LogicalVolume) Exists(ctx context.Context) (bool, error) {
 	}
 
 	for _, existingLV := range existingLVs {
-		if existingLV.Name == lv.Name {
+		if existingLV.Name == lv.Name && existingLV.VGName == lv.VolumeGroup.Name {
 			return true, nil
 		}
 	}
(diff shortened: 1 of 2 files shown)


bdevcich:
> If the block debice fails to activate

You have a typo in your commit message

pkg/blockdevice/lvm/logical_volumes.go
```diff
@@ -187,7 +187,7 @@ func (lv *LogicalVolume) Activate(ctx context.Context, rawArgs string) (bool, er
 		}
 	}
 
-	return false, fmt.Errorf("could not find logical volume %s: %w", lv.Name, err)
+	return false, fmt.Errorf("could not find logical volume %s", lv.Name)
```
bdevcich, line 190: Why drop the error detail here?

bdevcich (approved):

pkg/blockdevice/lvm/logical_volumes.go
matthew-richerson: `err` is nil here, so it was a mistake to include it originally. That's why we would see this error message:
"could not find logical volume lv-0: %!w(<nil>)"

roehrich-hpe (approved):

matthew-richerson:
> > If the block debice fails to activate
> 
> You have a typo in your commit message

I'll ammend the commit

PR #458, opened 2025-03-04
Run commands on MGT after Lustre targets are activated

matthew-richerson:
This commit adds the ability to run commands on an MGT after all the Lustre targets have been activated. This works for MGTs created with a workflow and MGTs from a pool.

The commands are specified in the storageprofile.data.lustreStorage.preMountMGTCommands field.

This commit also coordinates the creation of MDTs and OSTs to ensure that all MDTs are created first.


roehrich-hpe (approved):

internal/controller/nnf_lustre_mgt_controller.go
```diff
@@ -364,6 +377,81 @@ func (r *NnfLustreMGTReconciler) EraseOldFsName(nnfLustreMgt *nnfv1alpha6.NnfLus
 	return nil
 }
 
+// HandleNewClaims looks for any new claims in Spec.ClaimList and assigns them
```
ajfloeder, line 380:
```suggestion
// HandleNewCommands looks for any new claims in Spec.ClaimList and assigns them
```

internal/controller/nnf_lustre_mgt_controller.go
```diff
@@ -364,6 +377,81 @@ func (r *NnfLustreMGTReconciler) EraseOldFsName(nnfLustreMgt *nnfv1alpha6.NnfLus
...
+			}
+
+			return nil, nil
+		}
+	}
+
+	return nil, nil
+}
+
+// RemoveOldClaims removes any old entries from the Status.ClaimList and erases the fsname from
```
ajfloeder, line 417:
```suggestion
// RemoveOldCommands removes any old entries from the Status.ClaimList and erases the fsname from
```

internal/controller/nnf_storage_controller.go
```diff
@@ -915,6 +938,96 @@ func (r *NnfStorageReconciler) getFsName(ctx context.Context, nnfStorage *nnfv1a
...
+		return nil, dwsv1alpha3.NewResourceError("could not find pinned storage profile").WithError(err).WithFatal()
+	}
+
+	// Some tests don't fake out the nnfStorage completely
+	if len(nnfStorage.Spec.AllocationSets) == 0 {
+		return nil, nil
+	}
+
+	// If this NnfStorage is for a standalone MGT, then we don't need to run the shared MGT commands
+	if len(nnfStorage.Spec.AllocationSets) == 1 && nnfStorage.Spec.AllocationSets[0].Name == "mgt" {
```
bdevcich, line 953: Would it be useful to log this information?

internal/controller/nnf_storage_controller.go
```diff
@@ -915,6 +938,96 @@ func (r *NnfStorageReconciler) getFsName(ctx context.Context, nnfStorage *nnfv1a
...
+	}
+
+	nnfLustreMgt, err := r.getLustreMgt(ctx, nnfStorage)
+	if err != nil {
+		return nil, dwsv1alpha3.NewResourceError("could not get NnfLustreMGT for address: %s", nnfStorage.Status.MgsAddress).WithError(err)
+	}
+
+	// An NnfLustreMgt in the "nnf-system" namespace indicates an external MGT. Commands can't be run on external MGTs.
+	if nnfLustreMgt.GetNamespace() == "nnf-system" {
+		return nil, nil
```
bdevcich, line 964: Same here maybe? (re: log)

bdevcich (approved):

internal/controller/nnf_storage_controller.go
matthew-richerson: Not really. This will happen whenever you're creating a new MGT for the pool. It's expected, and the MGT commands are a no-op.

internal/controller/nnf_storage_controller.go
matthew-richerson: I can add a comment to this one in case anyone gets confused why their MGT commands aren't working for an external MGT. I think we're the only ones that run that way, though.

PR #459, opened 2025-03-06
Rescan devices on ClientMount deletion

matthew-richerson:
Rescan devices when the ClientMount resource is deleted to make sure we pick up the removal of the NVMe namespaces. Also, print a log message anytime the NVMe rescan results in a change to the number of NVMe namespaces in /dev.

### DIFF
--- a/pkg/blockdevice/nvme/nvme.go
+++ b/pkg/blockdevice/nvme/nvme.go
@@ -77,20 +77,84 @@ func NvmeListDevices(log logr.Logger) ([]NvmeDevice, error) {
 	return devices, nil
 }
 
-func NvmeRescanDevices(log logr.Logger) error {
+func NvmeGetDevices() ([]string, error) {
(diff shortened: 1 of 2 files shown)


pkg/blockdevice/nvme/nvme.go
```diff
@@ -77,20 +77,84 @@ func NvmeListDevices(log logr.Logger) ([]NvmeDevice, error) {
...
+				endingNamespaces = append(endingNamespaces[:i], endingNamespaces[i+1:]...)
+				break
 			}
 		}
+		if !found {
+			removedNamespaces = append(removedNamespaces, startingNamespace)
+		}
+	}
+
+	if len(removedNamespaces) != 0 {
```
ajfloeder, line 152: Thanks for putting this in Matt!

ajfloeder (approved):

PR #461, opened 2025-03-17
Copy Offload API: Support changes to using mpi-operator

bdevcich:
This adds a few pieces to support using mpi-operator for the new Copy Offload API.

Most notably is adding the `NNF_COPY_OFFLOAD_SERVER` environment variable which tells the compute nodes how to contact the Copy Offload Server that is hosting the API on one of the rabbit nodes allocated to the workflow.

### DIFF
--- a/internal/controller/nnf_workflow_controller_container_helpers.go
+++ b/internal/controller/nnf_workflow_controller_container_helpers.go
@@ -177,8 +177,8 @@ func (c *nnfUserContainer) createMPIJob() error {
 		return err
 	}
 
-	// Add the ports to the worker spec and add environment variable for both launcher/worker For
-	// copy offload we want the ports on the launcher so the copy offload server can be contacted
+	// Add the ports to the worker spec and add environment variable for both launcher/worker. For
+	// copy offload we want the ports on the launcher, so the copy offload server can be contacted
 	// from the computes.
 	// FIXME: For user-containers, we're opening the ports on the workers - but is that what we
 	// actually want?
--- a/internal/controller/nnf_workflow_controller_container_helpers.go
+++ b/internal/controller/nnf_workflow_controller_container_helpers.go
@@ -43,18 +43,19 @@ import (
 )
 
 type nnfUserContainer struct {
-	workflow *dwsv1alpha3.Workflow
-	profile  *nnfv1alpha6.NnfContainerProfile
-	nnfNodes []string
-	volumes  []nnfContainerVolume
-	secrets  []nnfContainerSecret
-	username string
-	uid, gid int64
-	client   client.Client
-	log      logr.Logger
-	scheme   *runtime.Scheme
-	ctx      context.Context
-	index    int
+	workflow    *dwsv1alpha3.Workflow
+	profile     *nnfv1alpha6.NnfContainerProfile
+	nnfNodes    []string
+	volumes     []nnfContainerVolume
+	secrets     []nnfContainerSecret
+	username    string
+	uid, gid    int64
+	client      client.Client
+	log         logr.Logger
+	scheme      *runtime.Scheme
+	ctx         context.Context
+	index       int
+	copyOffload bool
 }
 
 // This struct contains all the necessary information for mounting container storages
@@ -79,6 +80,7 @@ type nnfContainerSecret struct {
 
 const (
 	requiresContainerAuth = "container-auth"
+	requiresCopyOffload   = "copy-offload"
 )
 
 // MPI container workflow. In this model, we use mpi-operator to create an MPIJob, which creates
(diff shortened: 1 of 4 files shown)


internal/controller/nnf_workflow_controller_container_helpers.go
```diff
@@ -486,6 +497,8 @@ func addHostPorts(spec *corev1.PodSpec, ports []uint16) {
 	}
 
 	// Add the ports to the containers
+	// FIXME: this adds the same ports to all containers. Is that what we actually want? Doesn't
```
roehrich-hpe, line 500: I would expect each container to get its own port.

The /metrics endpoint in our controller-manager pods is not an apple-to-apples comparision, but it's close enough to give an idea of how ports work in multi-container pods:

- the controller-manager container listens on :8080, which has nothing that makes it visible to things from outside the pod.
- the kube-rbac-proxy container listens on :8443, which is marked as a "containerPort", so it's visible to things outside the pod.  This port has a Service resource; here's the one for nnf-controller-manager: `kubectl get svc -n nnf-system nnf-controller-manager-metrics-service`
- the kube-rbac-proxy process receives metrics requests from outside the pod on :8443 and does the rbac checks and then queries the /metrics endpoint at :8080, where it's talking to the controller-manager in the other container within the same pod.

internal/controller/nnf_workflow_controller_helpers.go
```diff
@@ -1535,6 +1540,21 @@ func (r *NnfWorkflowReconciler) waitForContainersToStart(ctx context.Context, wo
 			}
 		}
 
+		// If we're up and running and this is copy offload, then go and find which rabbit node the
```
roehrich-hpe, line 1543: "go and find" => "determine"

roehrich-hpe (approved):

internal/controller/nnf_workflow_controller_container_helpers.go
bdevcich: Agreed.

I discovered this when I was using multiple containers in the spec before I went back to mpi-operator. One container for the server, and one worker. I discovered that this logic tries to add the same port to both containers, which doesn't work.

internal/controller/nnf_workflow_controller_helpers.go
```diff
@@ -1535,6 +1540,21 @@ func (r *NnfWorkflowReconciler) waitForContainersToStart(ctx context.Context, wo
...
+		// the copy offload server.
+		if running && slices.Contains(workflow.Status.Requires, requiresCopyOffload) {
+			jobList, err := r.getMPIJobChildrenJobs(ctx, workflow, mpiJob)
+			if err != nil || len(jobList.Items) < 1 {
+				return nil, dwsv1alpha3.NewResourceError("could not retrieve MPIJob Child Jobs to find launcher node").WithError(err).WithFatal()
+			}
+
+			launcherJob := jobList.Items[0]
+			launcherNnfNode := launcherJob.Spec.Template.Spec.NodeSelector["kubernetes.io/hostname"]
+			workflow.Status.Env["NNF_COPY_OFFLOAD_SERVER"] = launcherNnfNode
```
roehrich-hpe, line 1554: So far, the plan is that the libcopyoffload library knows how to figure out which rabbit to talk to. Up to now, that means looking at a special file on the compute to read the rabbit's name.
With this PR, I think we want libcopyoffload to first look for this environment variable and use it, if found. If not found, then revert to looking for that special file.

roehrich-hpe: In that case, let's make this env var less specific to copy-offload. Let's name it "NNF_USER_SERVER".

bdevcich: In this particular case though, it's only set when `requires=copy-offload`.

roehrich-hpe: We probably need the env var any time the container profile uses an mpiJob.

bdevcich: Agreed. Perhaps we do both: Have a generic one for `NNF_CONTAINER_LAUNCHER` and also keep `NNF_COPY_OFFLOAD_SERVER` for this case to make it easy on the user.

But based on your comment above, maybe it doesn't need to be easy if the library defaults to having the environment variable.

roehrich-hpe: Just one, something that makes sense to someone writing an app for their user container.
For copy-offload, the library will know how to get it right and the user won't need to fuss with it.

internal/controller/nnf_workflow_controller_container_helpers.go
bdevcich: I'll address this in another PR

roehrich-hpe (approved):

internal/controller/nnf_workflow_controller_helpers.go
```diff
@@ -1535,9 +1540,21 @@ func (r *NnfWorkflowReconciler) waitForContainersToStart(ctx context.Context, wo
...
+		// set an env var for flux. This tells the computes where to contact the MPI Launcher
+		if running {
+			jobList, err := r.getMPIJobChildrenJobs(ctx, workflow, mpiJob)
+			if err != nil || len(jobList.Items) < 1 {
+				return nil, dwsv1alpha3.NewResourceError("could not retrieve MPIJob Child Jobs to find launcher node").WithError(err).WithFatal()
+			}
+
+			launcherJob := jobList.Items[0]
+			launcherNnfNode := launcherJob.Spec.Template.Spec.NodeSelector["kubernetes.io/hostname"]
+			workflow.Status.Env["NNF_CONTAINER_LAUNCHER"] = launcherNnfNode
```
roehrich-hpe, line 1553: I am updating my copy-offload doc PR  to add this.

bdevcich:
> Update the PR description with the new variable name.

I will do that when I squash.

bdevcich:
> > Update the PR description with the new variable name.
> 
> I will do that when I squash.

And someone how I messed that up 🤦

PR #467, opened 2025-03-24
Containers: Open numPorts host ports for each container

bdevcich:
Previously, the implementation did not account for multiple containers in a PodSpec, leading to issues where the same port was mapped across multiple containers, which is invalid.

This update ensures that numPorts ports are opened for each container in the PodSpec. For non-MPI containers, this applies to all containers. For MPI containers, only the Launcher containers are assigned ports, as the workers are accessed via mpirun through the Launcher.

The port numbers are exposed via the NNF_CONTAINER_PORTS environment variable. Additionally, new environment variables are created for each container using the container name, e.g.,
NNF_CONTAINER_PORTS_<container_name>.

This change also fixes issues related to supporting multiple containers and resolves a problem with the subdomain environment variable.

Examples of environment variables:
1. Single container with 1 port (numPorts=1):

```
NNF_CONTAINER_DOMAIN: default.svc.cluster.local
NNF_CONTAINER_HOSTNAMES: copy-offload-blake-launcher copy-offload-blake-worker-0 copy-offload-blake-worker-1
NNF_CONTAINER_LAUNCHER: rabbit-node-1
NNF_CONTAINER_PORTS: "5000"
NNF_CONTAINER_PORTS_copy_offload_srvr: "5000"
NNF_CONTAINER_SUBDOMAIN: copy-offload-blake
```

2. Two containers with 2 ports each (numPorts=2):

```
NNF_CONTAINER_DOMAIN: default.svc.cluster.local
NNF_CONTAINER_HOSTNAMES: rabbit-node-1 rabbit-node-2
NNF_CONTAINER_PORTS: 5000,5001,5002,5003
NNF_CONTAINER_PORTS_port_test_0: 5000,5001
NNF_CONTAINER_PORTS_port_test_1: 5002,5003
NNF_CONTAINER_SUBDOMAIN: test-blake
```

### DIFF
--- a/internal/controller/nnf_workflow_controller_container_helpers.go
+++ b/internal/controller/nnf_workflow_controller_container_helpers.go
@@ -712,13 +712,13 @@ func (c *nnfUserContainer) addEnvVars(workflow dwsv1alpha3.Workflow, spec *corev
 		// Add env variables to workflow
 		workflow.Status.Env["NNF_CONTAINER_SUBDOMAIN"] = subdomain
 		workflow.Status.Env["NNF_CONTAINER_DOMAIN"] = domain
-		workflow.Status.Env["NNF_CONTAINER_HOSTNAMES"] = strings.Join(hosts, " ")
+		workflow.Status.Env["NNF_CONTAINER_HOSTNAMES"] = strings.Join(hosts, ",")
 
 		// Add env variables to container
 		container.Env = append(container.Env,
 			corev1.EnvVar{Name: "NNF_CONTAINER_SUBDOMAIN", Value: subdomain},
 			corev1.EnvVar{Name: "NNF_CONTAINER_DOMAIN", Value: domain},
-			corev1.EnvVar{Name: "NNF_CONTAINER_HOSTNAMES", Value: strings.Join(hosts, " ")},
+			corev1.EnvVar{Name: "NNF_CONTAINER_HOSTNAMES", Value: strings.Join(hosts, ",")},
 			corev1.EnvVar{Name: "ENVIRONMENT", Value: os.Getenv("ENVIRONMENT")},
 			corev1.EnvVar{
 				Name: "NNF_NODE_NAME",
--- a/internal/controller/nnf_workflow_controller_container_helpers.go
+++ b/internal/controller/nnf_workflow_controller_container_helpers.go
@@ -171,29 +171,29 @@ func (c *nnfUserContainer) createMPIJob() error {
 	// passwd Init Container.
 	c.addInitContainerWorkerWait(launcherSpec, len(c.nnfNodes))
 
-	// Get the ports from the port manager
+	// Get the ports from the port manager and map them to each container
 	ports, err := c.getHostPorts()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/controller/nnf_workflow_controller_container_helpers.go
```diff
@@ -420,7 +427,7 @@ func (c *nnfUserContainer) applyPermissions(spec *corev1.PodSpec, mpiJobSpec *mp
 		container := &spec.Containers[idx]
 
 		// Add an InitContainer to map the user to the provided uid/gid using /etc/passwd
-		c.addInitContainerPasswd(spec, container.Image)
+		c.addInitContainerPasswd(spec, container.Image, idx)
```
roehrich-hpe, line 430: I don't get this. Wouldn't you need just one init container to create the /etc/passwd, and then mount that into all other containers?

internal/controller/nnf_workflow_controller_container_helpers.go
```diff
@@ -611,32 +686,35 @@ func (c *nnfUserContainer) addSecrets(spec *corev1.PodSpec) {
...
 			}
 		} else {
-			subdomain = spec.Subdomain
 			hosts = append(hosts, c.nnfNodes...)
 		}
 
+		// Add env variables to workflow
+		workflow.Status.Env["NNF_CONTAINER_SUBDOMAIN"] = subdomain
+		workflow.Status.Env["NNF_CONTAINER_DOMAIN"] = domain
+		workflow.Status.Env["NNF_CONTAINER_HOSTNAMES"] = strings.Join(hosts, " ")
```
roehrich-hpe, line 715: Why wouldn't this be comma-separated?

internal/controller/nnf_workflow_controller_helpers.go
```diff
@@ -2157,32 +2162,35 @@ func (r *NnfWorkflowReconciler) checkContainerPorts(ctx context.Context, workflo
...
+
+	pm, err := getContainerPortManager(ctx, r.Client)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, alloc := range pm.Status.Allocations {
+		if alloc.Requester != nil && alloc.Requester.UID == workflow.UID {
+			if alloc.Status == nnfv1alpha6.NnfPortManagerAllocationStatusInUse && len(alloc.Ports) == numPortsToRequest {
+				// // Add workflow env var for the ports
```
roehrich-hpe, line 2181: This PR already satisfies this, right? Can this comment be removed now?

internal/controller/nnf_workflow_controller_container_helpers.go
bdevcich: Great question. This should be commas.

internal/controller/nnf_workflow_controller_container_helpers.go
bdevcich: The way it works now is you get an InitContainer for every container and they cannot have the same names. What you're suggesting is an improvement and I will do that as a follow-up.

PR #472, opened 2025-03-26
Container profiles: Add copy offload checks

bdevcich:
Ensure that container names that indicate copy offload have the correct service account and also ensure the profile is using MPI.

### DIFF
--- a/api/v1alpha7/nnfcontainerprofile_webhook.go
+++ b/api/v1alpha7/nnfcontainerprofile_webhook.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"os"
 	"reflect"
+	"regexp"
 	"strings"
 
 	"github.com/kubeflow/mpi-operator/pkg/apis/kubeflow/v2beta1"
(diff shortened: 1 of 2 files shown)


api/v1alpha7/nnfcontainerprofile_webhook.go
```diff
@@ -135,6 +146,22 @@ func (r *NnfContainerProfile) validateContent() error {
 		if !workerOk || len(worker.Template.Spec.Containers) < 1 {
 			return fmt.Errorf("MPISpec.MPIReplicaSpecs.Worker must be present with at least 1 container defined")
 		}
+
+		// When it looks like we have a copy offload Launcher container, ensure the service account is correct
+		for _, c := range launcher.Template.Spec.Containers {
+			if isCopyOffloadContainer(c.Name) {
```
roehrich-hpe, line 152: if (it looks like it's not a copy-offload profile) {
  Hey, you don't get to use the copy-offload service account.
}

bdevcich: Good call. Done.

ajfloeder (approved):

PR #477, opened 2025-04-09
Allow deletion to make more progress when Computes and Rabbits are un…

matthew-richerson:
…responsive

- Ignore ClientMount resources that won't transition to "unmounted" for computes that are offline during PostRun
- Remove finalizers from ClientMounts during Teardown for computes that are offline
- Allow deletion of NnfNodeStorage and NnfNodeBlockStorages to proceed in parallel between different Rabbits

### DIFF
--- a/internal/controller/nnf_access_controller.go
+++ b/internal/controller/nnf_access_controller.go
@@ -49,6 +49,7 @@ import (
 	"github.com/DataWorkflowServices/dws/utils/updater"
 	nnfv1alpha7 "github.com/NearNodeFlash/nnf-sos/api/v1alpha7"
 	"github.com/NearNodeFlash/nnf-sos/internal/controller/metrics"
+	"github.com/NearNodeFlash/nnf-sos/pkg/helpers"
 )
 
 // NnfAccessReconciler reconciles a NnfAccess object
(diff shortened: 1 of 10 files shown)


internal/controller/nnf_access_controller.go
```diff
@@ -1149,6 +1161,137 @@ func clientMountName(access *nnfv1alpha7.NnfAccess) string {
...
+			}
+		}
+	}
+
+	return nil
+}
+
+// checkOfflineCompute finds the Storage resource for the Rabbit physically attached to the clientMount's
+// compute node, and checks whether the compute node is marked as "Offline" (indicating no PCIe connection)
+func (r *NnfAccessReconciler) checkOfflineCompute(ctx context.Context, nnfAccess *nnfv1alpha7.NnfAccess, clientMount *dwsv1alpha3.ClientMount) (bool, error) {
```
ajfloeder, line 1226: Can this method be made common for all controllers?

internal/controller/nnf_access_controller.go
```diff
@@ -1149,6 +1161,137 @@ func clientMountName(access *nnfv1alpha7.NnfAccess) string {
...
+			return compute.Status == dwsv1alpha3.OfflineStatus, nil
+		}
+	}
+
+	return false, nil
+}
+
+// getRabbitFromClientMount finds the name of the Rabbit that is physically connected to the ClientMount's
+// compute node
+func (r *NnfAccessReconciler) getRabbitFromClientMount(ctx context.Context, clientMount *dwsv1alpha3.ClientMount) (string, error) {
```
bdevcich, line 1262: Similar to Tony's comment - and I believe we talked about this during standup, but could this be made available too? nnf-dm needs to do something similar and it would be nice to have an "official" way to do this.

roehrich-hpe: During standup we talked about putting it under sos's pkg/ dir.

bdevcich (approved):

PR #478, opened 2025-04-09
Reduce CRD Size by Replacing core.PodSpecs with Slimmed Nnf* Versions

bdevcich:
The CRD size has grown too large due to the use of core.PodSpecs and
mpi-operator's MPIJobSpec, making it unwieldy and difficult to manage.

To address this, slimmed-down versions of PodSpec and Container have
been introduced, which are now used for the container profile's Spec and
MPISpec fields.

Additionally:
- A `copyOffload` flag has been added to the MPISpec to simplify
  defining the Copy Offload container profile by automating the correct
  service account usage.
- The `slotsPerWorker` field has been added to the NnfMPISpec to allow
  adjustment of slots in user container jobs. Copy Offload defers to the
  Data Movement profile's slot configuration.

Changes to the NnfDataMovementManager will follow in a subsequent PR.


api/v1alpha6/conversion_test.go
```diff
@@ -102,6 +110,125 @@ func TestFuzzyConversion(t *testing.T) {
...
+			containers[i].RestartPolicy = nil
+			containers[i].ResizePolicy = nil
+			containers[i].VolumeDevices = nil
+			containers[i].ReadinessProbe = nil
+			containers[i].StartupProbe = nil
+			containers[i].LivenessProbe = nil
+			containers[i].Lifecycle = nil
+			containers[i].TerminationMessagePath = ""
+			containers[i].TerminationMessagePolicy = ""
+			containers[i].ImagePullPolicy = ""
```
roehrich-hpe, line 143: Some of these we want, maybe in the next PR, definitely in the same release.
This one is already documented here: https://nearnodeflash.github.io/latest/guides/user-containers/readme/#using-a-private-container-repository

roehrich-hpe (approved):

api/v1alpha6/conversion_test.go
bdevcich: Looks like `imagePullSecrets` for sure?

api/v1alpha7/nnfcontainerprofile_types.go
```diff
@@ -69,33 +69,114 @@ type NnfContainerProfileData struct {
...
+
+	// MPIJobSpec to define the MPI containers created from this profile.
+	// Either this or NnfSpec must be provided, but not both.
+	// +kubebuilder:validation:Rule="(self.spec != null) != (self.mpiSpec != null)",Message="Exactly one of 'spec' or 'mpiSpec' must be set."
+	NnfMPISpec *NnfMPISpec `json:"mpiSpec,omitempty"`
+}
+
+// NnfPodSpec represents the specification of a pod that can be used in a container profile. This is
+// a slimmed down version of a corev1.PodSpec to reduce the size of the CRD.
+type NnfPodSpec struct {
```
roehrich-hpe, line 92: You're not representing the 'mpiReplicaSpecs' layer.

```
$ kubectl apply -f environments/kind/nnf-dm/reference/copy-offload-template-nnfcontainerprofile.yaml
Error from server (BadRequest): error when creating "environments/kind/nnf-dm/reference/copy-offload-template-nnfcontainerprofile.yaml": NnfContainerProfile in version "v1alpha7" cannot be handled as a NnfContainerProfile: strict decoding error: unknown field "data.mpiSpec.mpiReplicaSpecs", unknown field "data.mpiSpec.runPolicy"
```

We could make upgrades easier by representing that layer, so the customer doesn't have to edit these yaml files, and fix the indentation, and probably get it wrong the first time so they have to do it again.

api/v1alpha6/conversion_test.go
roehrich-hpe: We could support the probe fields (3, I count), and the TerminationMessage* fields.
I don't mind doing that in the next PR.

api/v1alpha7/nnfcontainerprofile_types.go
roehrich-hpe: Wow, never mind. The existing yamls need big changes anyway.

This wasn't caught by unpack-manifests.  I'll have to look at that.

api/v1alpha6/conversion_test.go
bdevcich: Yea, let's save that for a follow up PR

bdevcich: I added an issue for this: https://github.com/NearNodeFlash/NearNodeFlash.github.io/issues/260

PR #484, opened 2025-04-17
Improve ClientMount unmounts

matthew-richerson:
This commit fixes two issues:
 - Don't run the NVMe rescan on ClientMount delete for lustre file systems
 - Check that the NnfNodeBlockStorage has deleted its StorageGroups after unmount before considering unmount complete


internal/controller/nnf_access_controller.go
```diff
@@ -913,6 +938,41 @@ func (r *NnfAccessReconciler) getBlockStorageAccessStatus(ctx context.Context, a
...
+
+				if access.Spec.IgnoreOfflineComputes {
+					computeOffline := false
+					for _, compute := range storage.Status.Access.Computes {
+						if compute.Name != statusNodeName {
+							continue
+						}
+
+						if compute.Status == dwsv1alpha3.OfflineStatus {
+							computeOffline = true
```
roehrich-hpe, line 964: break out of the for-loop here ?

roehrich-hpe (approved):

PR #494, opened 2025-05-05
Keep updating NnfNodeBlockStorage /dev until NnfNodeStorage is deleted

matthew-richerson:
The NnfNodeBlockStorage and NnfNodeStorage resources both get marked for deletion at the same time. This can cause an issue if the nnf-node-manager restarts before the NnfNodeStorage is fully deleted. The NnfNodeStorage controller relies on the NnfNodeBlockStorage keeping the status section up to date with the latest /dev paths for the NVMe namespaces.

This change forces the NnfNodeBlockStorage controller to skip the deletion path code until the NnfNodeStorage has finished deleting.

### DIFF
--- a/internal/controller/nnf_node_block_storage_controller.go
+++ b/internal/controller/nnf_node_block_storage_controller.go
@@ -183,14 +183,11 @@ func (r *NnfNodeBlockStorageReconciler) Reconcile(ctx context.Context, req ctrl.
 	defer func() { err = statusUpdater.CloseWithStatusUpdate(ctx, r.Client.Status(), err) }()
 	defer func() { nodeBlockStorage.Status.SetResourceErrorAndLog(err, log) }()
 
-	if !nodeBlockStorage.GetDeletionTimestamp().IsZero() {
+	// If the NnfNodeStorage hasn't removed the finalizer from this NnfNodeBlockStorage, then don't start
+	// the deletion process. We still might need the /dev paths updated for the NnfNodeStorage to properly
+	// teardown
+	if !nodeBlockStorage.GetDeletionTimestamp().IsZero() && !controllerutil.ContainsFinalizer(nodeBlockStorage, finalizerNnfNodeStorage) {
 		if !controllerutil.ContainsFinalizer(nodeBlockStorage, finalizerNnfNodeBlockStorage) {
-			return ctrl.Result{}, nil
-		}
-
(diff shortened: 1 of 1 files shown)


internal/controller/nnf_node_block_storage_controller.go
```diff
@@ -183,17 +183,14 @@ func (r *NnfNodeBlockStorageReconciler) Reconcile(ctx context.Context, req ctrl.
 	defer func() { err = statusUpdater.CloseWithStatusUpdate(ctx, r.Client.Status(), err) }()
 	defer func() { nodeBlockStorage.Status.SetResourceErrorAndLog(err, log) }()
 
-	if !nodeBlockStorage.GetDeletionTimestamp().IsZero() {
+	// If the NnfNodeStorage hasn't removed the finalizer from this NnfNodeBlockStorage, then don't start
+	// the deletion process. We still might need the /dev paths updated for the NnfNodeStorage to properly
+	// teardown
+	if !nodeBlockStorage.GetDeletionTimestamp().IsZero() && !controllerutil.ContainsFinalizer(nodeBlockStorage, finalizerNnfNodeStorage) {
```
roehrich-hpe, line 189: If zero, and does have the finalizer, do you need to return with a retry?
You shouldn't go down to the non-deletion path.

matthew-richerson: Here's what I want:

| Has deletion Timestamp    |  Has NnfNodeStorage finalizer | Go down delete path |
| -------- | ------- |------|
| No  | No    | No |
| No  | Yes    | No |
| Yes    | No   | Yes |
| Yes | Yes| No |

In all cases where we don't go down the delete path, we need to go down the non-deletion path. The finalizer it's checking for is one that was put on by the NnfNodeStorage controller. We don't want the NnfNodeBlockStorage to start doing anything deletion related until the NnfNodeStorage controller removes that finalizer. And we need the NnfNodeBlockStorage to have updated /dev paths in the status section in the event of a pod restard. That's why we have to go down the non-deletion path in the last case in the table.

I'm having trouble figuring out which case you were asking about. Does that explanation explain it?

ajfloeder (approved):

internal/controller/nnf_node_block_storage_controller.go
roehrich-hpe: If the api server will allow spec or status updates on a resource that is marked for deletion, then carry on.

PR #496, opened 2025-05-07
checkContainersResults: account for SuccessCriteriaMet

bdevcich:


internal/controller/nnf_workflow_controller_helpers.go
```diff
@@ -1861,7 +1861,12 @@ func (r *NnfWorkflowReconciler) checkContainersResults(ctx context.Context, work
 
 		for _, job := range jobList.Items {
 			for _, condition := range job.Status.Conditions {
-				if condition.Type != batchv1.JobComplete {
+				switch condition.Type {
+				// Once the SuccessCriteriaMet condition is set, the job will eventually transition to JobComplete. Watch for both cases.
+				// TODO: At some point, we should upgrade the API to use batchv1.SuccessCriteriaMet
```
roehrich-hpe, line 1866: This is interesting if they've added to the API without bumping the API version. Maybe there are times when that can be done safely.

roehrich-hpe (approved):

PR #501, opened 2025-06-03
re-vendor before release

matthew-richerson:


roehrich-hpe (approved):
You should also vendor the new DWS API everywhere before any repo picks up any other changes.

PR #505, opened 2025-06-12
Add debug for disabled computes

matthew-richerson:
Add debug information when disabled computes are ignored. Also, add debug when deactivating the block devices fails.


pkg/blockdevice/nvme/nvme.go
```diff
@@ -133,6 +133,10 @@ func NvmeRescanDevices(log logr.Logger) error {
 		return err
 	}
 
+	if len(endingNamespaces) != 0 || len(endingNamespaces) != 16 {
+		log.Info("nvme ns-rescan found an unexpected number of namespaces", "count", len(endingNamespaces), "namespaces", endingNamespaces)
```
ajfloeder, line 137: We need to support less than 16 namespaces. It would be good to add the number you were expected here too, and we do need to support less than 16 somehow.

bdevcich (approved):

PR #531, opened 2025-09-05
Allow a longer grace time for renewing the leader-election lease.

roehrich-hpe:
Add some commandline args to the controller to control the LeaseDuration, RenewDeadline, and RetryPeriod. Adjust the patch that builds the manifest to insert these.


roehrich-hpe:
I have a gitops patch as well, so we can tune it that way if necessary.

PR #590, opened 2026-03-11
Add NnfContainerData resource

matthew-richerson:
Create a resource that gets created during PreRun to describe the mount paths available on the Rabbit for a job. This is driven by the "#dw container" directive to create the NnfAccesses to trigger the mounts on the Rabbit. The NnfContainerProfile now has a new option "createContainer" that can toggle between creating Pods to launch containers on the Rabbits, or to go through all the same steps but stop short of creating the actual pods and create the NnfContainerData resource instead.


ajfloeder (commented):
Good feature addition — the NnfContainerData CRD, conversion handling, and RBAC are all cleanly wired up. A few issues to address, one critical:

1. **Critical**: `waitForContainersToFinish` and `checkContainersResults` will error/hang in PostRun when `CreateContainer=false` (no Jobs/MPIJobs exist to check)
2. Orphan Service created for non-MPI containers when `CreateContainer=false`
3. Full Job/MPIJob spec built and discarded when `CreateContainer=false` — check could move earlier
4. `deleteContainers` uses individual deletes for NnfContainerData while MPIJob/Job use `DeleteAllOf`
5. Unused `nnfcontainerdatalog` logger variable in webhook file
6. `createContainer: true` in all examples is redundant given the kubebuilder default

roehrich-hpe:
That looks like a copilot-generated review.

ajfloeder (approved):

PR #610, opened 2026-04-24
Add nnf CLI for completing common tasks

matthew-richerson:
This commit adds a python script that can be used for automating common tasks on NNF resources. To start it has "create_persistent" and "destroy_persistent" subcommands.

The script is structured to allow additional subcommands to be added. It provides an interface for walking the workflow through all phases and providing hooks after each phase completes.


ajfloeder (commented):
*(Review retracted)*

tools/nnf/src/nnf/commands/create_persistent.py
```diff
@@ -0,0 +1,267 @@
...
+
+    print(f"Workflow '{workflow_name}' created.")
+
+    ok, msg = workflow.run_to_completion(wf, args.timeout)
+    if not ok:
+        print(f"error: {msg}")
+        return 2
+
+    print(f"Persistent storage '{args.name}' created successfully.")
+    return 0
```
ajfloeder: **Bug – unhandled exceptions orphan Workflows in the cluster.**

If `run_to_completion` or any hook raises an unexpected exception (e.g. a `TypeError`, `KeyError`, or anything that isn't `ApiException`/`ValueError`), the Workflow is left behind in the cluster with no cleanup.

Wrapping the create+run block in `try/finally` would guarantee `teardown_and_delete` runs:

```python
try:
    k8s.create_object(...)
except kubernetes.client.exceptions.ApiException as exc:
    ...
    return 2

print(f"Workflow '{workflow_name}' created.")

try:
    ok, msg = workflow.run_to_completion(wf, args.timeout)
    if not ok:
        print(f"error: {msg}")
        return 2
except Exception:
    workflow.teardown_and_delete(wf.name, wf.namespace, args.timeout)
    raise

print(f"Persistent storage '{args.name}' created successfully.")
return 0
```

Same pattern applies to `destroy_persistent.py`.

tools/nnf/src/nnf/workflow.py
```diff
@@ -0,0 +1,754 @@
...
+                namespace=namespace,
+                plural=crd.DWS_WORKFLOW_PLURAL,
+                name=workflow_name,
+            )
+        except kubernetes.client.exceptions.ApiException as exc:
+            return False, f"API error while polling: {exc.reason} (HTTP {exc.status})"
+
+        status: Dict[str, Any] = obj.get("status", {})
+        current_state = status.get("state", "")
+        ready = status.get("ready", False)
```
ajfloeder, line 139: **Resilience – a single transient API error aborts the entire workflow.**

If the k8s API returns a transient 503/429 during polling, `wait_for_state` immediately returns `(False, ...)` and the entire workflow is torn down. For a tool that may run against a busy cluster, retrying a few times on 5xx would be more robust:

```python
except kubernetes.client.exceptions.ApiException as exc:
    if exc.status is not None and exc.status >= 500:
        LOGGER.info("Transient API error (%s), retrying...", exc.status)
        time.sleep(POLL_INTERVAL)
        continue
    return False, f"API error while polling: {exc.reason} (HTTP {exc.status})"
```

A similar concern applies to `_wait_for_breakdown`.

tools/nnf/src/nnf/commands/create_persistent.py
```diff
@@ -0,0 +1,267 @@
...
+    parser: argparse.ArgumentParser = add_command_parser(
+        subparsers,
+        "create_persistent",
+        help="Create a persistent storage instance.",
+    )
+    parser.add_argument(
+        "--name",
+        required=True,
+        help="Name of the persistent storage instance.",
+    )
```
ajfloeder, line 38: **Usability – `--name` is not validated against K8s naming rules.**

The name is embedded directly into Workflow resource names (`nnf-create-persistent-{name}`), which must conform to DNS subdomain rules (lowercase alphanumeric, hyphens, max 253 chars). An invalid name like `My_Storage!` produces a confusing K8s API error.

Consider validating at the CLI boundary before hitting the cluster:

```python
import re

_DNS_SUBDOMAIN_RE = re.compile(r'^[a-z0-9]([a-z0-9\-]*[a-z0-9])?$')

def _validate_name(name: str) -> bool:
    full = f"nnf-create-persistent-{name}"
    return len(full) <= 253 and bool(_DNS_SUBDOMAIN_RE.match(full))
```

This would also apply to `destroy_persistent`.

tools/nnf/src/nnf/workflow.py
```diff
@@ -0,0 +1,754 @@
+"""Shared helpers for creating, driving, and populating DWS Workflow resources.
```
ajfloeder, line 1: **Nit / architecture – consider splitting this file.**

At 754 lines, `workflow.py` handles four distinct concerns:
1. Workflow state machine (advance, wait, run_to_completion, teardown)
2. Servers/Computes population (fill_servers, fill_servers_default, fill_computes)
3. Allocation-set math (build_alloc_sets, _scale_to_count)
4. Resource helpers (get_storage_profile, has_standalone_mgt, get_rabbits_from_system_config)

Moving #2–#4 into a `servers.py` (or similar) would keep each module focused and make the already-good test structure map more naturally to the source layout. Not blocking, but would help as more subcommands are added.

tools/nnf/pyproject.toml
```diff
@@ -0,0 +1,24 @@
...
+requires = ["setuptools>=42"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "nnf"
+version = "0.1.0"
+requires-python = ">=3.6"
+dependencies = [
+    "kubernetes>=12.0,<27",
+]
```
ajfloeder: **Bug – `kubernetes` upper bound is too wide for Python 3.6.**

`kubernetes>=12.0,<27` allows pip to resolve versions 24+ which dropped Python 3.6 support. On a 3.6 system this will either fail at install time (if the wheel metadata declares `python_requires >= 3.8`) or at import time.

Cap to the last 3.6-compatible release:

```toml
dependencies = [
    "kubernetes>=12.0,<24",
]
```

tools/nnf/tests/test_workflow.py
```diff
@@ -0,0 +1,365 @@
...
+    with patch("nnf.workflow.advance", MagicMock()), \
+            patch("nnf.workflow.wait_for_state", return_value=(True, "")), \
+            patch("nnf.workflow.teardown_and_delete", mock_teardown):
+        ok, msg = run_to_completion(
+            _make_wf(state_hooks={"Proposal": [hook]}), timeout=60
+        )
+
+    assert ok is False
+    assert "not enough rabbits" in msg
+    mock_teardown.assert_called_once()
```
ajfloeder, line 360: **Testing gap – `teardown_and_delete` failure paths are not covered.**

Only the happy path (advance succeeds → wait succeeds → delete) is tested. Missing scenarios:
- `advance` to Teardown raises `ApiException` → should still call `delete`
- `wait_for_state("Teardown")` returns `(False, ...)` → should still call `delete`

These are important because `teardown_and_delete` is the safety net for every failure in `run_to_completion`.

tools/nnf/src/nnf/commands/create_persistent.py
```diff
@@ -0,0 +1,267 @@
...
+        default=None,
+        help="Storage profile name to pass as profile= in the #DW directive (optional).",
+    )
+    parser.add_argument(
+        "--timeout",
+        type=int,
+        default=workflow.DEFAULT_TIMEOUT,
+        help=f"Seconds to wait for each workflow state (default: {workflow.DEFAULT_TIMEOUT}).",
+    )
+    parser.set_defaults(func=run)
```
ajfloeder: **Nit – `--alloc-count` argparse help text is narrower than the actual behavior.**

The argparse help says "for the AllocatePerCompute strategy" but `build_alloc_sets` also uses it for `AllocateAcrossServers`. Interestingly, the README's description ("controls how many allocations are made on each Rabbit node") is more accurate. Consider updating the argparse help to match the README wording.

tools/nnf/README.md
```diff
@@ -0,0 +1,178 @@
...
+## create_persistent
+
+Show help for the `create_persistent` subcommand:
+
+```bash
+nnf create_persistent --help
+```
+
+Use `--profile` to select the `NnfStorageProfile` if the default profile is not desired. For Lustre, the profile drives allocation-set behavior for labels such as `ost`, `mdt`, `mgt`, and `mgtmdt`.
+
```
ajfloeder, line 110: **Nit – supported `--fs-type` choices aren't listed in the README.**

Examples only show `xfs` and `lustre`. A one-liner like "Supported filesystem types: `raw`, `xfs`, `gfs2`, `lustre`" would save readers a trip to `--help`.

tools/nnf/README.md
```diff
@@ -0,0 +1,178 @@
...
+```bash
+nnf --verbose create_persistent --help
+nnf create_persistent --verbose --help
+```
+
+## create_persistent
+
+Show help for the `create_persistent` subcommand:
+
+```bash
```
ajfloeder, line 89: **Nit – several useful flags are undocumented.**

`--namespace`, `--user-id`, `--group-id`, and `--timeout` exist on both subcommands but aren't mentioned here. `--help` covers them, but `--timeout` in particular seems worth calling out since the default is 180s and a slow cluster might need more.

ajfloeder (commented):
Seems like a cleanup function for orphaned workflows would be handy.

tools/nnf/pyproject.toml
```diff
@@ -0,0 +1,24 @@
+[build-system]
+requires = ["setuptools>=42"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "nnf"
+version = "0.1.0"
+requires-python = ">=3.6"
+dependencies = [
+    "kubernetes>=12.0,<24",
```
ajfloeder, line 10: **`requires-python = ">=3.6"` — Python 3.6 has been EOL since Dec 2021.**

The code itself is fine on 3.6 (f-strings, typing), but 3.6 signals to users that ancient interpreters are supported when they likely aren't tested. Consider bumping to `>=3.9` (or whatever the minimum version that's actually available in your deployment environments). This also unlocks `dict` union operators and `list`/`dict` as generic types if desired later.

tools/nnf/src/nnf/servers.py
```diff
@@ -0,0 +1,484 @@
...
+            f"Workflow '{workflow_name}'."
+        )
+
+    bd_name: str = bd_ref["name"]
+    bd_namespace: str = bd_ref.get("namespace", namespace)
+
+    ok, msg = _wait_for_breakdown(bd_name, bd_namespace, timeout)
+    if not ok:
+        return False, msg
+
```
ajfloeder, line 359: **`fill_computes` is defined but unused.**

This function isn't called by any subcommand in this PR. It's reasonable to include it as forward-looking infrastructure, but it means it's untested against real workflows. Consider either:

- Adding it as part of a command that uses it (e.g. `jobdw`), or
- Dropping it for now and adding it when a consumer arrives

If keeping it, a brief docstring note like "Used by future `jobdw`/`persistdw` subcommands" would help readers understand why it's here.

tools/nnf/src/nnf/commands/destroy_persistent.py
```diff
@@ -0,0 +1,111 @@
...
+        "--group-id",
+        type=int,
+        default=None,
+        dest="group_id",
+        help="Group ID that owns this storage (default: current group).",
+    )
+    parser.add_argument(
+        "--timeout",
+        type=int,
+        default=workflow.DEFAULT_TIMEOUT,
```
ajfloeder: **Reduce boilerplate: extract shared workflow arguments and create→run→teardown pattern.**

`--namespace`, `--user-id`, `--group-id`, and `--timeout` are defined identically in both `create_persistent` and `destroy_persistent`. Consider a shared `add_workflow_arguments(parser)` helper in `commands/__init__.py` alongside the existing `add_common_arguments`.

Additionally, the create→run→teardown error-handling flow at the bottom of both `run()` functions is nearly identical:

```python
try:
    k8s.create_object(...)
except ApiException:
    ...
    return 2
try:
    ok, msg = workflow.run_to_completion(wf, args.timeout)
    if not ok:
        ...
        return 2
except Exception:
    workflow.teardown_and_delete(...)
    raise
```

Extracting this into e.g. `workflow.create_and_run(wf, args)` would reduce duplication and make new commands less error-prone to write.

tools/nnf/src/nnf/__init__.py
```diff
@@ -0,0 +1,62 @@
...
+        ),
+    )
+    add_common_arguments(parser)
+
+    subparsers = parser.add_subparsers(dest="command", metavar="<command>")
+    subparsers.required = True
+
+    create_persistent.register(subparsers)
+    destroy_persistent.register(subparsers)
+
```
ajfloeder: **Extensibility: consider auto-discovery for subcommands.**

Registration is hardcoded here — every new subcommand requires touching this file. If the tool is "expected to grow" (as the README says), consider entry-point-based or directory-scanning discovery, e.g. iterate `nnf.commands.*` modules that expose a `register()` function:

```python
import importlib, pkgutil
from nnf import commands as _cmds_pkg

for _info in pkgutil.iter_modules(_cmds_pkg.__path__):
    mod = importlib.import_module(f"nnf.commands.{_info.name}")
    if hasattr(mod, "register"):
        mod.register(subparsers)
```

With only 2 commands this is fine to leave as-is, but flagging it now so the pattern is established early if more are coming soon.

tools/nnf/src/nnf/commands/create_persistent.py
```diff
@@ -0,0 +1,283 @@
...
+            if val is not None
+        ]
+        if conflicts:
+            print(f"error: --rabbit-count cannot be combined with: {', '.join(conflicts)}")
+            return 1
+        if args.rabbit_count < 1:
+            print("error: --rabbit-count must be at least 1")
+            return 1
+        try:
+            all_rabbits = servers.get_rabbits_from_system_config()
```
ajfloeder, line 171: **Nit: log the randomly chosen rabbits.**

`random.sample` here makes debugging harder since the selection isn't visible unless the user digs into `fill_servers_default` logs. Consider adding an `LOGGER.info("Selected rabbits: %s", rabbits)` right after this line so `--verbose` output always shows which rabbits were picked.

tools/nnf/src/nnf/crd.py
```diff
@@ -0,0 +1,32 @@
+"""CRD group, version, and plural constants for NNF and DWS resources."""
+
+# ---------------------------------------------------------------------------
+# NNF (nnf.cray.hpe.com)
+# ---------------------------------------------------------------------------
+
+NNF_GROUP = "nnf.cray.hpe.com"
+NNF_VERSION = "v1alpha11"
```
ajfloeder, line 9: **Hardcoded CRD versions will need manual updates on every version bump.**

`NNF_VERSION` and `DWS_VERSION` are pinned here. These will drift as CRD versions advance. Consider either:

1. Making them overridable via CLI flags or env vars (e.g. `NNF_CLI_NNF_VERSION`), or
2. Querying the preferred version from the API server at runtime (`/apis/<group>`)

At minimum, a comment here noting these need to be kept in sync with the CRD definitions would help future maintainers.

tools/nnf/src/nnf/workflow.py
```diff
@@ -0,0 +1,266 @@
...
+                "jobID": 0,
+                "userID": self.user_id,
+                "groupID": self.group_id,
+                "wlmID": "nnf-cli",
+            },
+        }
+
+
+def advance(workflow_name: str, namespace: str, state: str) -> None:
+    """Patch the Workflow's ``spec.desiredState`` to *state*."""
```
ajfloeder, line 95: **`WorkflowRun` conflates manifest building with runtime hook configuration.**

This class builds the k8s manifest (via `.manifest`) *and* carries runtime state (`state_hooks`, `on_state_complete`). If future subcommands need to customize the manifest structure (different labels, annotations, finalizers, owner references), those concerns will be mixed with hook management.

For now this is manageable, but as the tool grows, consider separating the manifest builder (what gets submitted to k8s) from the run configuration (how the state machine is driven). For example, `manifest` could be a plain dict parameter rather than a property derived from the class fields.

ajfloeder (commented):
## Architecture & Extensibility Review

Nice work — the overall architecture is solid. The hook-based state machine in `WorkflowRun`, the clean module decomposition (`k8s`, `workflow`, `servers`, `profile`, `crd`, `utils`), and the `register(subparsers)` pattern all set up well for extensibility. The test coverage is thorough (~1,570 lines across 7 test files).

The inline comments focus on preparing for growth:

1. **Subcommand auto-discovery** — avoid touching `__init__.py` for every new command
2. **Shared workflow arguments + create→run→teardown helper** — the two commands duplicate ~20 lines of arg definitions and ~15 lines of error-handling boilerplate
3. **Hardcoded CRD versions** — will need manual updates on every version bump
4. **`fill_computes` is defined but unused** — forward-looking, but worth noting
5. **`requires-python >= 3.6`** — consider bumping to a supported version
6. **Log randomly chosen rabbits** — small debuggability improvement
7. **`WorkflowRun` dual responsibility** — manifest builder + hook container; minor concern at current scale

None of these are blockers for merging — they're suggestions for making the next 5 commands easier to add.

tools/nnf/pyproject.toml
roehrich-hpe: It was my understanding that they requested that Matt use 3.6. Is this a case where you didn't proof-read the LLM's review comment, or didn't give it sufficient guidance?

ajfloeder (approved):

PR #612, opened 2026-04-29
Add Rabbit Node commands

matthew-richerson:
This commit adds some more commands to the nnf tool to manage Rabbit nodes. Specifically, it adds: drain, undrain, disable, enable, and df.

This commit also changes the format of the persistent create and destroy commands to allow better grouping of commands. The format is now: nnf persistent create/destroy ...

### DIFF
--- a/tools/nnf/README.md
+++ b/tools/nnf/README.md
@@ -2,7 +2,7 @@
 
 This directory contains a small Python helper CLI for working with NNF and DWS resources from the `nnf-sos` repository.
 
-Today the tool focuses on helper workflows around persistent storage. It is expected to grow over time with additional subcommands, so this document is the local reference point for setup, usage, and examples.
+Today the tool focuses on helper workflows around persistent storage and Rabbit node management. It is expected to grow over time with additional subcommands, so this document is the local reference point for setup, usage, and examples.
 
 ## Scope
 
@@ -11,6 +11,9 @@ The `nnf` command is intended as a repository-local helper tool, not a standalon
 - create a persistent storage instance for testing
 - destroy a persistent storage instance
 - inspect and drive workflow-based storage operations from a simple CLI
+- drain or undrain Rabbit nodes to control workflow scheduling
+- disable or enable Rabbit nodes to control workload manager allocations
+- check disk space usage on Rabbit nodes
 
 The command operates against an existing Kubernetes environment where the relevant NNF and DWS CRDs are already installed.
 
@@ -57,15 +60,16 @@ nnf --help
 Show command-specific help:
 
 ```bash
-nnf create_persistent --help
-nnf destroy_persistent --help
+nnf persistent create --help
+nnf persistent destroy --help
+nnf rabbit --help
 ```
 
 Enable verbose workflow logging:
 
 ```bash
-nnf --verbose create_persistent --help
-nnf create_persistent --verbose --help
+nnf --verbose persistent create --help
+nnf persistent create --verbose --help
 ```
 
 Both subcommands support these common flags:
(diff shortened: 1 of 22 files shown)


tools/nnf/README.md
```diff
@@ -200,7 +204,101 @@ Key `destroy_persistent` flags:
...
+nnf rabbit drain rabbit-node-0
+```
+
+Drain multiple nodes with a reason:
+
+```bash
+nnf rabbit drain rabbit-node-0 rabbit-node-1 --reason "firmware update"
+```
+
+If the storage annotation fails after the node taint succeeds, the CLI attempts to roll back the taint. If the rollback also fails, the CLI logs the error and continues to the next node.
```
ajfloeder: The code annotates storage first, then taints. If the taint fails, it rolls back the annotation. The README describes the inverse. The README needs to say: "If the node taint fails after the storage annotation succeeds, the CLI attempts to roll back the annotation."

tools/nnf/src/nnf/commands/rabbit/undrain.py
```diff
@@ -0,0 +1,58 @@
...
+        metavar="NODE",
+        help="One or more Rabbit node names to undrain.",
+    )
+    parser.set_defaults(func=run)
+
+
+def run(args: argparse.Namespace) -> int:
+    """Execute the rabbit undrain sub-command."""
+
+    def action(node_name: str) -> int:
```
ajfloeder, line 46: If _remove_drain_taints fails, the storage annotations are already gone but the taint still exists. The node appears undrained in the storage metadata but is still tainted in Kubernetes. drain has a rollback for this exact pattern; undrain doesn't.

tools/nnf/src/nnf/k8s.py
```diff
@@ -128,3 +129,84 @@ def delete_object(
...
+        "stdout": True,
+        "tty": False,
+    }
+    if container is not None:
+        kwargs["container"] = container
+    raw: str = kubernetes.stream.stream(
+        api.connect_get_namespaced_pod_exec, **kwargs
+    )
+    if strip_channel_bytes:
+        raw = raw.lstrip("\x00\x01\x02\x03")
```
ajfloeder: `str.lstrip` treats its argument as a **set of characters** and strips all consecutive leading characters that belong to that set — not just one. A response whose JSON payload happened to start with `\x01` (e.g. `\x01\x01{...}`) would have both bytes stripped, which is fine here, but a payload that legitimately began with `\x00` would also be silently truncated.

Since the WebSocket framing is a single prefix byte, stripping exactly one character is the right intent:

```python
if strip_channel_bytes and raw and raw[0] in "\x00\x01\x02\x03":
    raw = raw[1:]
```

Or with a regex for clarity:
```python
import re
if strip_channel_bytes:
    raw = re.sub(r'^[\x00-\x03]', '', raw)
```

tools/nnf/src/nnf/commands/rabbit/df.py
```diff
@@ -0,0 +1,169 @@
...
+    for rabbit in rabbits:
+        storage = storage_by_name.get(rabbit)
+        if storage is None:
+            print(f"error: no Storage resource found for '{rabbit}'", file=sys.stderr)
+            errors += 1
+            continue
+
+        if not _is_enabled_ready(storage):
+            skipped.append(rabbit)
+            continue
```
ajfloeder, line 152: When the user explicitly names a node (e.g. `nnf rabbit df rabbit-node-X`), a disabled or not-ready node silently falls through to the `skipped` summary footer rather than producing an error. A user who specifies a node by name most likely expects either data or an explicit error, not a quiet footnote.

Suggested fix — distinguish explicit requests from the default "all nodes" sweep:

```python
if not _is_enabled_ready(storage):
    if args.nodes:
        # user explicitly requested this node — treat it as an error
        spec_state = storage.get("spec", {}).get("state", "")
        status_state = storage.get("status", {}).get("status", "")
        print(
            f"error: '{rabbit}' is not Enabled/Ready "
            f"(spec.state={spec_state}, status={status_state})",
            file=sys.stderr,
        )
        errors += 1
    else:
        skipped.append(rabbit)
    continue
```

tools/nnf/src/nnf/commands/rabbit/undrain.py
```diff
@@ -0,0 +1,58 @@
...
+import sys
+from typing import Any, Dict, List
+
+import kubernetes.client.exceptions  # type: ignore[import-untyped]
+
+from nnf import crd
+from nnf import k8s
+from nnf.commands import add_command_parser
+from nnf.commands.rabbit._helpers import ERROR, OK, for_each_node
+from nnf.commands.rabbit.drain import TAINT_KEY, _remove_drain_annotations, _remove_drain_taints
```
ajfloeder: Seems like these private helpers should either be promoted to public, or moved to the helpers.

ajfloeder (commented):
Check the timestamps throughout. Seems like local time would be more useful than UTC.

ajfloeder (approved):

PR #615, opened 2026-05-05
Add more subcommands to nnf script

matthew-richerson:
This commit adds the flowschema, state, and version subcommands. It also adds new share and unshare subcommands for allowing persistent storage to be shared between multiple users.

The destroy persistent command was changed to automatically create a workflow with a UID that matches the UID of the persistent storage being destroyed.

### DIFF
--- a/tools/nnf/src/nnf/commands/system/flowschema.py
+++ b/tools/nnf/src/nnf/commands/system/flowschema.py
@@ -141,7 +141,7 @@ def _list_priority_levels(api_version: str) -> int:
     return 0
 
 
-def _view_activity(api_version: str, cred_name: str) -> int:
+def _view_activity(api_version: str, flow_schema_name: str) -> int:
     """View metrics for a specific FlowSchema."""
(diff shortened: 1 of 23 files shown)


tools/nnf/src/nnf/commands/system/flowschema.py
```diff
@@ -0,0 +1,205 @@
...
+            queue_len = str(queuing.get("queueLengthLimit", "")) if queuing else ""
+        else:
+            shares = queues = hand_size = queue_len = ""
+        rows.append((name, type_val, shares, queues, hand_size, queue_len))
+
+    _print_table(
+        ("NAME", "TYPE", "NOMINALCONCURRENCYSHARES", "QUEUES", "HANDSIZE", "QUEUELENGTHLIMIT"),
+        rows,
+    )
+    return 0
```
ajfloeder, line 141: Nit: `cred_name` is misleading here — this is a flow schema name, not a credential. Consider renaming to `flow_schema_name` to match the argparse dest (`args.flow_schema_name`).

ajfloeder (approved):

PR #622, opened 2026-05-08
rabbit_state.sh: add k8s state script

behlendorf:
Add Dean's rabbit_state.sh script which can be used to summarize the state of the NNF deployment.  It would be desirable to convert this to a `nnf system state` sub-command.  I thought this was already in one of the repo's but I wasn't able to find it.  Regardless, I think it would be useful to include it here.

### DIFF
--- a/scripts/rabbit_state.sh
+++ b/scripts/rabbit_state.sh
@@ -1,0 +1,254 @@
+#!/bin/bash
+#
+# Output a status summary of the NNF software deployed to the Kubernetes cluster.
+#
+
+set -euo pipefail
+
+for cmd in kubectl jq nodeattr awk paste grep; do
+    command -v "$cmd" >/dev/null || { echo "error: $cmd not found" >&2; exit 1; }
+done
+
+export KUBECONFIG=/etc/kubernetes/admin.conf
+export CLUSTER=$(nodeattr -q cluster | awk -F'[' '{print $1}')
+[[ -n $CLUSTER ]] || { echo "error: could not determine cluster name" >&2; exit 1; }
+echo "CLUSTER=$CLUSTER"
+
+export VERSION=$(kubectl get deploy -n nnf-system nnf-controller-manager -o json | jq -rM .metadata.labels | awk -F'"' '/nnf-version/ { print $4 }')
+echo "VERSION=$VERSION"
+
+FMT="+%Y%m%d%H%M%S"
+now=$(date $FMT)
+echo "now=$now"
+histdir="$TMPDIR/rabbit-hist-$CLUSTER"
+if [[ ! -d $histdir ]]; then
+  mkdir "$histdir" || exit 1
+fi
+if [[ -f $histdir/latest-stamp ]]; then
+  prevstamp=$(<"$histdir/latest-stamp")
+  echo "prevstamp=$prevstamp"
+fi
+
+kubectl get pods -n nnf-system -o wide > "$histdir/pods-$now"
+kubectl get pods -n dws-system -o wide > "$histdir/pods-dws-$now"
+kubectl get pods -n nnf-dm-system -o wide > "$histdir/pods-dm-$now"
+kubectl get pods -n calico-system -o wide > "$histdir/pods-calico-$now"
+
+kubectl get nodes -o wide > "$histdir/nodes-$now"
+
+
+kubectl get nodes -o json | jq -cM '.items[] | {"name":.metadata.name,"taints":.spec.taints}' > "$histdir/taints-$now"
+
+kubectl get systemconfiguration default -o yaml > $histdir/systemconfiguration-"$now"
+
+kubectl get lustrefilesystems -A -o yaml > $histdir/lustrefilesystems-"$now"
+
+(
+
+  # DWS is expected to be on the worker nodes.
+  # Find these with: nodeattr -q rabbitk8sworkerinfra
+  worker_nodes=$(nodeattr -c rabbitk8sworkerinfra | tr ',' '|')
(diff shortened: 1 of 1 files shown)


scripts/rabbit_state.sh
```diff
@@ -0,0 +1,245 @@
+#!/bin/bash
```
ajfloeder, line 1: Missing `set -euo pipefail` after the shebang. Without it, all the top-level `kubectl ... > $histdir/...` snapshot calls fail silently — if `kubectl` can't reach the cluster or a namespace doesn't exist, you get empty files and the rest of the analysis produces misleading output (zero counts, no taints, etc.).

```bash
#!/bin/bash
set -euo pipefail
```

scripts/rabbit_state.sh
```diff
@@ -0,0 +1,245 @@
...
+#
+# Output a status summary of the NNF software deployed to the Kubernetes cluster.
+#
+
+export KUBECONFIG=/etc/kubernetes/admin.conf
+export CLUSTER=$(nodeattr -q cluster | awk -F'[' '{print $1}')
```
ajfloeder, line 13: `$CLUSTER` is never validated. If `nodeattr` isn't found or returns no value, this will be empty and `histdir` becomes `/var/tmp/root/rabbit-hist-` — every cluster's history lands in the same directory. Add a guard after the assignment:

```bash
export CLUSTER=$(nodeattr -q cluster | awk -F'[' '{print $1}')
[[ -n $CLUSTER ]] || { echo "error: could not determine cluster name" >&2; exit 1; }
```

scripts/rabbit_state.sh
```diff
@@ -0,0 +1,245 @@
...
+    local pending
+
+    local others
+
+    not_running=$(grep -c -v -e NAME -e Running "$f")
+    container_creating=$(grep -c ContainerCreating "$f")
+    crashloopbackoff=$(grep -c CrashLoopBackOff "$f")
+    createcontainererror=$(grep -c CreateContainerError "$f")
+    errimagepull=$(grep -c ErrImagePull "$f")
+    imagepullbackoff=$(grep -c ImagePullBackOff "$f")
```
ajfloeder: `errimagepull` and `imagepullbackoff` are missing `local` declarations, unlike every other status variable in this function. They'll persist in the subshell scope across subsequent calls to `summarize_pods`, causing stale counts to bleed into later namespaces if `grep -c` finds no matches and leaves the previous value set.

Add them to the existing `local` block above:

```bash
    local not_running
    local container_creating
    local terminating
    local crashloopbackoff
    local createcontainererror
    local errimagepull
    local imagepullbackoff
    local pending
```

scripts/rabbit_state.sh
```diff
@@ -0,0 +1,245 @@
...
+    if (( nodes_notready > 0 )); then
+      echo "      $nodes_notready are on NotReady nodes"
+    fi
+    if (( nodes_ready > 0 )); then
+      echo "      $nodes_ready are on Ready nodes"
+      echo "$nodes_ready_pods"
+      echo
+    fi
+  }
+
```
ajfloeder, line 135: Both `count_node_state` and `count_node_state_simple` write to the same hardcoded path `$histdir/count_node_state`. This is fine sequentially today, but is fragile — if these are ever called in parallel, they'll clobber each other's input. Use `mktemp` and clean up with a `trap`:

```bash
statefile=$(mktemp)
trap 'rm -f "$statefile"' RETURN
```

scripts/rabbit_state.sh
ajfloeder, line 13: No preflight check for required tools. `kubectl`, `jq`, `nodeattr`, `awk`, `paste`, and `tee` are all hard dependencies — if any are missing the script will fail mid-run with cryptic errors rather than a clear message up front. `nodeattr` in particular is HPE-cluster-specific and won't exist everywhere. Consider adding a guard early in the script:

```bash
for cmd in kubectl jq nodeattr; do
    command -v "$cmd" >/dev/null || { echo "error: $cmd not found" >&2; exit 1; }
done
```

scripts/rabbit_state.sh
```diff
@@ -0,0 +1,245 @@
...
+    echo
+    echo "DWS on unexpected nodes:"
+    echo "$ERR_NODES" | paste -d, -s
+    echo
+  fi
+
+  # The worker nodes should not be tainted.
+  # Find these with: nodeattr -q rabbitk8sworkerinfra
+  #UNEXP_TAINTS=$(grep -E '"elcap2[456]"' $histdir/taints-"$now" | grep -v null)
+  pat='"'"($worker_nodes)"'"'
```
ajfloeder, line 61: These commented-out lines with hardcoded `elcap2[456]` cluster names are leftover test-cluster artifacts. The active code already uses `$worker_nodes` dynamically, so these dead comments just add noise. Please remove them before merging.

scripts/rabbit_state.sh
```diff
@@ -0,0 +1,245 @@
...
+    echo
+    echo "Unexpected taints on nodes:"
+    echo "$UNEXP_TAINTS"
+    echo
+  fi
+
+  # Any node other than the worker nodes should be tainted for the NLCs. These
+  # are the rabbit nodes.
+  # Find these with: nodeattr -q rabbit
+  #NULL_TAINTS=$(grep null $histdir/taints-"$now" | grep -v -E '"elcap2[456]"')
```
ajfloeder: Same as above — hardcoded `elcap2[456]` test-cluster artifact. Please remove.

ajfloeder (changes requested):

ajfloeder (approved):
Squashing the commits would be great, thanks.

PR #623, opened 2026-05-08
nnf: add man pages for nnf utility

behlendorf:
Add a high-level `nnf(1)` man page as well an individual pages for each available sub-command: `nnf-persistent(1)`, `nnf-rabbit(1)`, and `nnf-system(1)`.

The `nnf-rabbit(1)` page includes the 'state' action which is currently part of the nnf-system command but a PR exists to move it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -512,3 +512,18 @@ catalog-build: opm ## Build a catalog image.
 .PHONY: catalog-push
 catalog-push: ## Push a catalog image.
 	$(MAKE) docker-push IMG=$(CATALOG_IMG)
+
+PREFIX ?= /usr/local
+MANDIR ?= $(PREFIX)/share/man
+MAN1DIR ?= $(MANDIR)/man1
+
+MAN1PAGES = man/man1/nnf.1 man/man1/nnf-persistent.1 man/man1/nnf-rabbit.1 man/man1/nnf-system.1
+
+.PHONY: install-man uninstall-man
+
+install-man:
+	mkdir -p $(DESTDIR)$(MAN1DIR)
+	install -m 0644 $(MAN1PAGES) $(DESTDIR)$(MAN1DIR)/
+
+uninstall-man:
+	rm -f $(addprefix $(DESTDIR)$(MAN1DIR)/,$(notdir $(MAN1PAGES)))
--- a/man/man1/nnf-rabbit.1
+++ b/man/man1/nnf-rabbit.1
@@ -1,0 +1,167 @@
+.TH NNF-RABBIT 1 "2026-05-08" "nnf 0.1.0" "User Commands"
+.SH NAME
+nnf-rabbit \- manage Rabbit node state
+.SH SYNOPSIS
+.B nnf rabbit
+.I action
+.RI [ options ]
+
+.PP
+.B nnf rabbit state
+.RI [ -v | --verbose ]
+
+.PP
+.B nnf rabbit disable
+.RI [ -r " REASON" ]
+.I NODE
+.RI [ NODE " ..." ]
+.RI [ -v | --verbose ]
+
+.PP
+.B nnf rabbit drain
+.RI [ -r " REASON" ]
+.I NODE
+.RI [ NODE " ..." ]
+.RI [ -v | --verbose ]
+
+.PP
+.B nnf rabbit enable
+.I NODE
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -512,3 +512,18 @@ catalog-build: opm ## Build a catalog image.
...
+MAN1PAGES = man/man1/nnf.1 man/man1/nnf-persistent.1 man/man1/nnf-rabbit.1 man/man1/nnf-system.1
+
+.PHONY: install-man uninstall-man
+
+install-man:
+	mkdir -p $(DESTDIR)$(MAN1DIR)
+	install -m 0644 $(MAN1PAGES) $(DESTDIR)$(MAN1DIR)/
+
+uninstall-man:
+	rm -f $(addprefix $(DESTDIR)$(MAN1DIR)/,$(MAN1PAGES))
```
ajfloeder: Bug: `addprefix` prepends the full relative path from `MAN1PAGES`, producing doubled paths at uninstall time, e.g.:

```
/usr/local/share/man/man1/man/man1/nnf.1
```

`install-man` avoids this because `install(1)` with a directory target implicitly uses the basename. `uninstall-man` needs `notdir` to strip the directory prefix first:

```makefile
uninstall-man:
	rm -f $(addprefix $(DESTDIR)$(MAN1DIR)/,$(notdir $(MAN1PAGES)))
```

man/man1/nnf-rabbit.1
```diff
@@ -0,0 +1,169 @@
...
+One or more requested operations failed, or resource queries failed.
+
+.SH EXAMPLES
+.TP
+Show Rabbit state summary:
+.B nnf rabbit state
+
+.TP
+Disable a Rabbit with a reason:
+.B nnf rabbit disable --reason 'maintenance rabbit-0'
```
ajfloeder: Bug: `rabbit-0` is being swallowed into the `--reason` string, leaving the required positional `NODE` argument missing. Running this example as-written would fail with an argparse error.

The `--reason` flag takes a single token; the node name must follow as a separate positional argument:

```
nnf rabbit disable --reason maintenance rabbit-0
```

Or, if the reason must contain spaces, the node name still needs to come after the quoted reason closes:

```
nnf rabbit disable --reason 'scheduled maintenance' rabbit-0
```

man/man1/nnf-rabbit.1
```diff
@@ -0,0 +1,169 @@
...
+This action displays a summary of:
+.IP \(bu 2
+NnfNode server health and status
+.IP \(bu 2
+DWS Storage state and status
+.IP \(bu 2
+Disabled and drained Rabbit nodes, including dates and reasons
+.IP \(bu 2
+Compute nodes associated with Rabbits that are marked with the
+badrabbit property.  These computes node will not be allocated by
```
ajfloeder:
```suggestion
badrabbit property.  These compute nodes will not be allocated by
```

man/man1/nnf-rabbit.1
```diff
@@ -0,0 +1,169 @@
...
+
+.PP
+.B nnf rabbit undrain
+.I NODE
+.RI [ NODE " ..." ]
+.RI [ -v | --verbose ]
+
+.SH DESCRIPTION
+.B nnf rabbit
+manages Rabbit storage node state.
```
ajfloeder: delete this line?

ajfloeder (changes requested):
**Pre-existing issue (not introduced by this PR):** `tools/nnf/src/nnf/commands/system/df.py` has the wrong module docstring — it reads `"""rabbit df sub-command:...` but should say `system df`. Not a blocker for this PR, but worth a follow-up.

Makefile
behlendorf: Done.

man/man1/nnf-rabbit.1
behlendorf: Fixed, updated to the properly quoted 'scheduled maintenance' example.

man/man1/nnf-rabbit.1
behlendorf: Fixed.

man/man1/nnf-rabbit.1
behlendorf: Removed.

ajfloeder (approved):

